PR #12, opened 2016-07-21
Infrastructure and preliminary layouts

JunTaoLuo:
While we are still working out the design, I'm adding the infrastructure (for build/test) of this repo. Much of the design and implementation will likely change so I'm keeping most of @Tratcher's initial work and layout for now. 

A few things I've noticed:
- Conflict with `ResponseCacheAttribute`:

This repo is likely going to be confused with the `ResponseCacheAttribute` in MVC: (https://github.com/aspnet/Mvc/blob/ee2cfa1963ee7f8195130d41330f94d4622ca89a/src/Microsoft.AspNetCore.Mvc.Core/ResponseCacheAttribute.cs). We should find a way to distinguish the two since the attribute just operates on the headers and the middleware here is described as `OutputCaching` in the docs (https://docs.asp.net/en/latest/performance/caching/output.html) and will actually operate as a cache. Also, if we choose to have the middleware to actively set the headers, how will this middleware interact with the MVC attribute? Will the two components exist side by side somehow? 

- Naming Cache vs Caching

In the initial work included in the branch, the naming is inconsistent and there are instances of `ResponseCaching*` and `ResponseCache*` and this should be standardized. Since the repo is constructed as `ResponseCachING` I started using that for now but I think `ResponseCache*` feels better API wise. For example, `app.UseResponseCache()` sounds more natural than `app.UseResponseCaching()`.

cc @DamianEdwards @davidfowl @muratg


BrennanConroy:
Xml comments needed?

BrennanConroy:
:shipit:

JunTaoLuo:
I don't think we'll have xml docs until the design is finalized

Tratcher:
:shipit:

PR #6, opened 2016-04-24
Update for AspNetCore

benaadams:
version 0.1.0

/cc @Tratcher

PR #13, opened 2016-07-27
Date & Age handling

benaadams:
Date & Age handling from #6

/cc @JunTaoLuo

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/IResponseCachingOptions.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/IResponseCachingOptions.cs
@@ -1,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    interface IResponseCachingOptions
+    {
+        int MaxCachedItemBytes { get; set; }
+    }
+}
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
@@ -2,16 +2,20 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Globalization;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.ResponseCaching
 {
     public class ResponseCachingContext
     {
         private string _cacheKey;
+        private RequestType _requestType;
 
         public ResponseCachingContext(HttpContext httpContext, IMemoryCache cache)
         {
@@ -31,12 +35,24 @@ namespace Microsoft.AspNetCore.ResponseCaching
 
         private bool CacheResponse { get; set; }
 
+        private bool IsProxied { get; set; }
+
         public bool CheckRequestAllowsCaching()
         {
             // Verify the method
             // TODO: What other methods should be supported?
-            if (!string.Equals("GET", HttpContext.Request.Method, StringComparison.OrdinalIgnoreCase))
+            if (string.Equals("GET", HttpContext.Request.Method, StringComparison.OrdinalIgnoreCase))
             {
+                _requestType = RequestType.FullReponse;
+            }
+            else if (string.Equals("HEAD", HttpContext.Request.Method, StringComparison.OrdinalIgnoreCase) ||
+                     string.Equals("OPTIONS", HttpContext.Request.Method, StringComparison.OrdinalIgnoreCase))
+            {
+                _requestType = RequestType.HeadersOnly;
+            }
+            else
+            {
+                _requestType = RequestType.NotCached;
                 return false;
             }
 
@@ -82,14 +98,24 @@ namespace Microsoft.AspNetCore.ResponseCaching
                     response.Headers[pair.Key] = pair.Value;
                 }
 
-                // TODO: Update cache headers (Age)
-                response.Headers["Served_From_Cache"] = DateTime.Now.ToString();
+                // TODO: Allow setting proxied _isProxied 
+                var age = Math.Max((DateTimeOffset.UtcNow - cacheEntry.Created).TotalSeconds, 0.0);
+                var ageString = (age > int.MaxValue ? int.MaxValue : (int)age).ToString(CultureInfo.InvariantCulture);
+                response.Headers[IsProxied ? "Age" : "X-Cache-Age"] = ageString;
 
-                // Copy the cached response body
-                var body = cacheEntry.Body;
-                if (body.Length > 0)
+                if (_requestType == RequestType.HeadersOnly)
                 {
-                    await response.Body.WriteAsync(body, 0, body.Length);
+                    response.Headers["Content-Length"] = "0";
+                }
+                else
+                {
+                    // Copy the cached response body
+                    var body = cacheEntry.Body;
+                    response.Headers["Content-Length"] = body.Length.ToString(CultureInfo.InvariantCulture);
+                    if (body.Length > 0)
+                    {
+                        await response.Body.WriteAsync(body, 0, body.Length);
+                    }
                 }
                 return true;
             }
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.ResponseCaching/IResponseCachingOptions.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    interface IResponseCachingOptions
```
Tratcher: Unused?

Class

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -82,14 +98,24 @@ private string CreateCacheKey()
...
+                var age = Math.Max((DateTimeOffset.UtcNow - cacheEntry.Created).TotalSeconds, 0.0);
+                var ageString = (age > int.MaxValue ? int.MaxValue : (int)age).ToString(CultureInfo.InvariantCulture);
+                response.Headers[IsProxied ? "Age" : "X-Cache-Age"] = ageString;
 
-                // Copy the cached response body
-                var body = cacheEntry.Body;
-                if (body.Length > 0)
+                if (_requestType == RequestType.HeadersOnly)
+                {
+                    response.Headers["Content-Length"] = "0";
```
benaadams: Content-Length should be correct size for HEAD?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -82,14 +98,24 @@ private string CreateCacheKey()
...
+                }
+                else
                 {
-                    await response.Body.WriteAsync(body, 0, body.Length);
+                    // Copy the cached response body
+                    var body = cacheEntry.Body;
+                    response.Headers["Content-Length"] = body.Length.ToString(CultureInfo.InvariantCulture);
+                    if (body.Length > 0)
+                    {
+                        await response.Body.WriteAsync(body, 0, body.Length);
```
vazzang: Should this call have a `ConfigureAwait(false)` on the end of the await?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -82,14 +98,24 @@ private string CreateCacheKey()
                     response.Headers[pair.Key] = pair.Value;
                 }
 
-                // TODO: Update cache headers (Age)
-                response.Headers["Served_From_Cache"] = DateTime.Now.ToString();
+                // TODO: Allow setting proxied _isProxied 
+                var age = Math.Max((DateTimeOffset.UtcNow - cacheEntry.Created).TotalSeconds, 0.0);
+                var ageString = (age > int.MaxValue ? int.MaxValue : (int)age).ToString(CultureInfo.InvariantCulture);
+                response.Headers[IsProxied ? "Age" : "X-Cache-Age"] = ageString;
```
JunTaoLuo: What is `IsProxied` for? What behaviour changes should this configure? Also, We should always be setting the `Age` header.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -118,15 +144,37 @@ internal bool OnResponseStarting()
...
+            // Don't cache errors? 404 etc
+            if (CacheResponse && HttpContext.Response.StatusCode == 200)
             {
                 // Store the buffer to cache
-                var cacheEntry = new ResponseCachingEntry();
-                cacheEntry.StatusCode = HttpContext.Response.StatusCode;
-                foreach (var pair in HttpContext.Response.Headers)
+                var cacheEntry = new ResponseCachingEntry()
+                {
+                    Created = DateTimeOffset.UtcNow,
```
JunTaoLuo: This should be set to the `Date` header from the response? That header contains the information on when the response was actually generated.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -118,15 +144,37 @@ internal bool OnResponseStarting()
...
+                    StatusCode = HttpContext.Response.StatusCode
+                };
+
+                var headers = HttpContext.Response.Headers;
+                var count = headers.Count
+                    - (headers.ContainsKey("Date") ? 1 : 0)
+                    - (headers.ContainsKey("Content-Length") ? 1 : 0)
+                    - (headers.ContainsKey("Age") ? 1 : 0);
+                var cachedHeaders = new List<KeyValuePair<string, StringValues>>(count);
```
JunTaoLuo: Hmm the values stored in here is just thrown away later on. Also, you can just set the values directly on `cacheEntry.Headers` now.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -118,15 +144,37 @@ internal bool OnResponseStarting()
...
+                var age = 0;
+                foreach (var entry in headers)
                 {
-                    cacheEntry.Headers[pair.Key] = pair.Value;
+                    // Reduce create date by Age 
+                    if (entry.Key == "Age" && int.TryParse(entry.Value, out age) && age > 0)
```
JunTaoLuo: I don't think we should rely on the `Age` header from upstream caches.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -118,15 +144,37 @@ internal bool OnResponseStarting()
...
+                    {
+                        cacheEntry.Created -= new TimeSpan(0, 0, age);
+                    }
+                    // Don't copy Date header or Content-Length
+                    else if (entry.Key != "Date" && entry.Key != "Content-Length")
```
JunTaoLuo: I think we should use the original date as provided by the origin. @Tratcher?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -31,12 +35,24 @@ public ResponseCachingContext(HttpContext httpContext, IMemoryCache cache)
...
         {
             // Verify the method
             // TODO: What other methods should be supported?
-            if (!string.Equals("GET", HttpContext.Request.Method, StringComparison.OrdinalIgnoreCase))
+            if (string.Equals("GET", HttpContext.Request.Method, StringComparison.OrdinalIgnoreCase))
+            {
+                _requestType = RequestType.FullReponse;
+            }
+            else if (string.Equals("HEAD", HttpContext.Request.Method, StringComparison.OrdinalIgnoreCase) ||
+                     string.Equals("OPTIONS", HttpContext.Request.Method, StringComparison.OrdinalIgnoreCase))
```
JunTaoLuo: RFC says "Responses to the OPTIONS method are not cacheable."

JunTaoLuo:
This is a good start and I have this change rolled in with some more changes I made to implement a basic HTTP cache for which I will open a PR soon (I'm keeping your contribution in the commit history). There are still a few implementation details that I would like clarification though.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
benaadams: It changes what Age means; e.g. if you are using ResponseCaching with https://github.com/aspnet/Proxy haven't quite worked out how they would communicate though

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
benaadams: Is when the cache entry is created?

JunTaoLuo: UtcNow at this point more accurately represents when the response was received by the cache, not when it was originally generated. The `Date` header is also what the RFC indicate to use for age calculations.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: From the RFC, we should always set `Age`. We might want to set the `X-Cache` headers but even that would be pending further discussion @Tratcher?

JunTaoLuo:
@benaadams thanks for the contribution! Your changes have been merged in https://github.com/aspnet/ResponseCaching/commit/7d716d20072b50ea7c6f9e56cc31ed204de1cd85

PR #14, opened 2016-07-28
ResponseCache updates to support varyby rules

JunTaoLuo:
Updating the design of ResponseCache/ResponseCacheEntries to support varyby rules. The design of how to configure these rules and what elements to vary by have not been finalized yet.

cc @Tratcher @davidfowl

### DIFF
--- a/samples/ResponseCachingSample/Startup.cs
+++ b/samples/ResponseCachingSample/Startup.cs
@@ -15,7 +15,7 @@ namespace ResponseCachingSample
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddMemoryCache();
+            services.AddDistributedResponseCache();
         }
 
         public void Configure(IApplicationBuilder app)
@@ -23,11 +23,14 @@ namespace ResponseCachingSample
             app.UseResponseCaching();
             app.Run(async (context) =>
             {
+                // These settings should be configured by context.Response.Cache.*
                 context.Response.GetTypedHeaders().CacheControl = new CacheControlHeaderValue()
                 {
                     Public = true,
-                    MaxAge = TimeSpan.FromSeconds(10)                    
+                    MaxAge = TimeSpan.FromSeconds(10)
                 };
+                context.Response.Headers["Vary"] = new string[] { "Accept-Encoding", "Non-Existent" };
+
                 await context.Response.WriteAsync("Hello World! " + DateTime.UtcNow);
             });
         }
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
@@ -3,9 +3,11 @@
 
 using System;
 using System.IO;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.Caching.Memory;
+using Microsoft.AspNetCore.ResponseCaching.Internal;
+using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.ResponseCaching
 {
@@ -13,15 +15,25 @@ namespace Microsoft.AspNetCore.ResponseCaching
     {
         private string _cacheKey;
 
-        public ResponseCachingContext(HttpContext httpContext, IMemoryCache cache)
+        public ResponseCachingContext(HttpContext httpContext, IResponseCache cache)
         {
+            if (cache == null)
+            {
+                throw new ArgumentNullException(nameof(cache));
+            }
+
+            if (httpContext == null)
+            {
+                throw new ArgumentNullException(nameof(httpContext));
+            }
+
             HttpContext = httpContext;
             Cache = cache;
         }
 
         private HttpContext HttpContext { get; }
 
-        private IMemoryCache Cache { get; }
+        private IResponseCache Cache { get; }
 
         private Stream OriginalResponseStream { get; set; }
 
@@ -46,38 +58,81 @@ namespace Microsoft.AspNetCore.ResponseCaching
         }
 
         // Only QueryString is treated as case sensitive
-        // GET;HTTP://MYDOMAIN.COM:80/PATHBASE/PATH?QueryString
+        // GET;/PATH;VaryBy
         private string CreateCacheKey()
         {
+            return CreateCacheKey(varyBy: null);
+        }
+
+        private string CreateCacheKey(CachedVaryBy varyBy)
+        {
             var request = HttpContext.Request;
-            return request.Method.ToUpperInvariant()
-                + ";"
-                + request.Scheme.ToUpperInvariant()
-                + "://"
-                + request.Host.Value.ToUpperInvariant()
-                + request.PathBase.Value.ToUpperInvariant()
-                + request.Path.Value.ToUpperInvariant()
-                + request.QueryString;
+            var builder = new StringBuilder()
+                .Append(request.Method.ToUpperInvariant())
+                .Append(";")
+                .Append(request.Path.Value.ToUpperInvariant())
+                .Append(CreateVaryByCacheKey(varyBy));
+
+            return builder.ToString();
+        }
+
+        private string CreateVaryByCacheKey(CachedVaryBy varyBy)
+        {
+            // TODO: resolve key format and delimiters
+            if (varyBy == null)
+            {
+                return string.Empty;
+            }
+
+            var request = HttpContext.Request;
+            var builder = new StringBuilder(";");
+
+            foreach (var header in varyBy.Headers)
+            {
+                var value = request.Headers[header].ToString();
+                // null vs Empty?
+                if (string.IsNullOrEmpty(value))
+                {
+                    value = "null";
+                }
+
+                builder.Append(header)
+                    .Append("=")
+                    .Append(value)
+                    .Append(";");
+            }
+
+            // Parse querystring params
+
+            return builder.ToString();
         }
 
         internal async Task<bool> TryServeFromCacheAsync()
         {
             _cacheKey = CreateCacheKey();
-            ResponseCachingEntry cacheEntry;
-            if (Cache.TryGetValue(_cacheKey, out cacheEntry))
+            var cacheEntry = Cache.Get(_cacheKey);
+
+            if (cacheEntry is CachedVaryBy)
+            {
+                // Request contains VaryBy rules, recompute key and try again
+                _cacheKey = CreateCacheKey(cacheEntry as CachedVaryBy);
+                cacheEntry = Cache.Get(_cacheKey);
+            }
+
+            if (cacheEntry is CachedResponse)
             {
                 // TODO: Compare cached request headers
-
-                // TODO: Evaluate Vary-By and select the most appropriate response
 
                 // TODO: Content negotiation if there are multiple cached response formats?
 
                 // TODO: Verify content freshness, or else re-validate the data?
 
+                var cachedResponse = cacheEntry as CachedResponse;
+
                 var response = HttpContext.Response;
                 // Copy the cached status code and response headers
-                response.StatusCode = cacheEntry.StatusCode;
-                foreach (var pair in cacheEntry.Headers)
+                response.StatusCode = cachedResponse.StatusCode;
+                foreach (var pair in cachedResponse.Headers)
                 {
                     response.Headers[pair.Key] = pair.Value;
                 }
@@ -86,7 +141,7 @@ namespace Microsoft.AspNetCore.ResponseCaching
                 response.Headers["Served_From_Cache"] = DateTime.Now.ToString();
 
                 // Copy the cached response body
-                var body = cacheEntry.Body;
+                var body = cachedResponse.Body;
                 if (body.Length > 0)
                 {
                     await response.Body.WriteAsync(body, 0, body.Length);
@@ -120,15 +175,34 @@ namespace Microsoft.AspNetCore.ResponseCaching
         {
             if (CacheResponse)
             {
-                // Store the buffer to cache
-                var cacheEntry = new ResponseCachingEntry();
-                cacheEntry.StatusCode = HttpContext.Response.StatusCode;
+                var response = HttpContext.Response;
+                var varyHeaderValue = response.Headers["Vary"];
+
+                // Check if any VaryBy rules exist
+                if (!StringValues.IsNullOrEmpty(varyHeaderValue))
+                {
+                    var cachedVaryBy = new CachedVaryBy
+                    {
+                        // Only vary by headers for now
+                        Headers = varyHeaderValue
+                    };
+
+                    Cache.Set(_cacheKey, cachedVaryBy);
+                    _cacheKey = CreateCacheKey(cachedVaryBy);
+                }
+
+                // Store the response to cache
+                var cachedResponse = new CachedResponse
+                {
+                    StatusCode = HttpContext.Response.StatusCode,
+                    Body = Buffer.ToArray()
+                };
                 foreach (var pair in HttpContext.Response.Headers)
                 {
-                    cacheEntry.Headers[pair.Key] = pair.Value;
+                    cachedResponse.Headers[pair.Key] = pair.Value;
                 }
-                cacheEntry.Body = Buffer.ToArray();
-                Cache.Set(_cacheKey, cacheEntry); // TODO: Timeouts
+
+                Cache.Set(_cacheKey, cachedResponse); // TODO: Timeouts
             }
 
             // TODO: TEMP, flush the buffer to the client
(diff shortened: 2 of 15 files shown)


src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/DistributedResponseCache.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public void Remove(string key)
+        {
+            _cache.Remove(key);
+        }
+
+        public void Set(string key, IResponseCacheEntry entry)
+        {
+            _cache.Set(key, DefaultResponseCacheSerializer.Serialize(entry));
```
JunTaoLuo: null checks needed to be added everywhere

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/IResponseCacheEntry.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    public interface IResponseCacheEntry
+    {
+        string Key { get; set; }
+        ResponseCacheEntryType Type { get; }
```
JunTaoLuo: Not necessarily needed since you can check the types by `<var> is <Type>` and serialize/deserialize with a special value.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -46,38 +46,88 @@ public bool CheckRequestAllowsCaching()
         }
 
         // Only QueryString is treated as case sensitive
-        // GET;HTTP://MYDOMAIN.COM:80/PATHBASE/PATH?QueryString
+        // GET;HTTP://MYDOMAIN.COM:80/PATHBASE/PATH?QueryString;VaryBy
```
JunTaoLuo: We'll need to finalize how keys are generated. Custom VaryBy, default components, etc.

samples/ResponseCachingSample/Startup.cs
```diff
@@ -15,19 +15,22 @@ public class Startup
...
             app.UseResponseCaching();
             app.Run(async (context) =>
             {
+                // These settings should be configured by context.Response.Cache.*
                 context.Response.GetTypedHeaders().CacheControl = new CacheControlHeaderValue()
                 {
                     Public = true,
                     MaxAge = TimeSpan.FromSeconds(10)                    
                 };
+                context.Response.GetTypedHeaders().AppendList("Vary", new string[] { "Accept-Encoding", "Non-Existent" });
```
Tratcher: ???
context.Response.Headers["Vary"] = new string[] { "Accept-Encoding", "Non-Existent" };

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/CachedResponse.cs
```diff
@@ -5,10 +5,16 @@
 
 namespace Microsoft.AspNetCore.ResponseCaching
 {
-    internal class ResponseCachingEntry
+    public class CachedResponse : IResponseCacheEntry
     {
+        public string Key { get; set; }
+
+        public ResponseCacheEntryType Type => ResponseCacheEntryType.Response;
```
Tratcher: Avoid the word Type, it's already defined by Object.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/DistributedResponseCache.cs
Tratcher: And hardening. None of these operations should throw.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/IResponseCacheEntry.cs
Tratcher: Remove

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/IResponseCacheEntry.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    public interface IResponseCacheEntry
+    {
+        string Key { get; set; }
```
Tratcher: Do you even need the key here? Isn't the key always given with the Get or Set?

JunTaoLuo: These keys can be long. We might choose to store by a hash of it inside the cache. I'll leave it in for now with a TODO to remove it if not needed.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/CachedVaryBy.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    public class CachedVaryBy : IResponseCacheEntry
```
Tratcher: Did your research show any platforms supporting multiple vary-by conditions?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/IResponseCache.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    public interface IResponseCache
+    {
+        IResponseCacheEntry Get(string key);
+        void Set(string key, IResponseCacheEntry entry);
```
Tratcher: Would the implementation be simpler or harder if we had separate methods for the cache entry and the vary-by entry? Then you wouldn't need IResponseCacheEntry or downcasting.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/CachedVaryBy.cs
JunTaoLuo: The RFC defines the `Vary` header as a list of other header values to vary by, and this is what the default caches of Apache, Django, RoR support but there are extensions such as node-cache that provide support for vary by params on node. ASP.NET supported a lot more like headers, params, content-encodings and a custom string.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/IResponseCache.cs
JunTaoLuo: Both would work, but not sure about the complexity. I imagine there might be common data between the two like key, expiry information, tags, etc.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: Seems like most implementations do not consider scheme, host and port in the key generation. I think that might be a good default to start with.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/IResponseCache.cs
JunTaoLuo: On second thought I'll remove `IResponseCacheEntry` until we have a proven reason to need them.

JunTaoLuo:
:up::date:

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/CachedResponse.cs
```diff
@@ -5,10 +5,12 @@
 
 namespace Microsoft.AspNetCore.ResponseCaching
 {
-    internal class ResponseCachingEntry
+    public class CachedResponse
     {
         public int StatusCode { get; set; }
+
         internal IHeaderDictionary Headers { get; set; } = new HeaderDictionary();
```
davidfowl: Why are these internal and status code public?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/DefaultResponseCacheEntrySerializer.cs
```diff
@@ -0,0 +1,136 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    public static class DefaultResponseCacheSerializer
```
davidfowl: Is the payload format documented in comments anywhere?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -120,15 +176,33 @@ internal void FinalizeCaching()
         {
             if (CacheResponse)
             {
-                // Store the buffer to cache
-                var cacheEntry = new ResponseCachingEntry();
-                cacheEntry.StatusCode = HttpContext.Response.StatusCode;
+                var response = HttpContext.Response;
+
+                // Check if any VaryBy rules exist
+                if (!string.IsNullOrEmpty(response.Headers["Vary"]))
```
davidfowl:
```C#
var varyHeaderValue = response.Headers["Vary"];

if (StringValues.IsNullOrEmpty(varyHeaderValue))
{
}

```

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/CachedResponse.cs
JunTaoLuo: No particular reason, could make everything internal until there's a reason to make them public. I don't see why these would be used by anyone else other than response caching middleware.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/DefaultResponseCacheEntrySerializer.cs
JunTaoLuo: Nope. Also I expect the format and payload to change as we progress. Maybe document it after we consolidate the design?

JunTaoLuo: Added some comments on the format.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/DistributedResponseCache.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Caching.Distributed;
+
+namespace Microsoft.AspNetCore.ResponseCaching
```
Tratcher: Internal?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/MemoryResponseCache.cs
```diff
@@ -0,0 +1,60 @@
...
+            _cache = cache;
+        }
+
+        public object Get(string key)
+        {
+            try
+            {
+                return _cache.Get(key);
+            }
+            catch
```
Tratcher: try/catch on MemoryCache?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -46,38 +57,84 @@ public bool CheckRequestAllowsCaching()
...
-                + request.Scheme.ToUpperInvariant()
-                + "://"
-                + request.Host.Value.ToUpperInvariant()
-                + request.PathBase.Value.ToUpperInvariant()
-                + request.Path.Value.ToUpperInvariant()
-                + request.QueryString;
+            var builder = new StringBuilder()
+                .Append(request.Method.ToUpperInvariant())
+                .Append(";")
+                .Append(request.PathBase.Value.ToUpperInvariant())
```
Tratcher: Do we even need PathBase?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -46,38 +57,84 @@ public bool CheckRequestAllowsCaching()
...
+        }
+
+        private string CreateVaryByCacheKey(CachedVaryBy varyBy)
+        {
+            // TODO: resolve key format and delimiters
+            if (varyBy == null)
+            {
+                return string.Empty;
+            }
+            else
```
Tratcher: The else is redundant after a return.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    public static class ResponseCachingServiceCollectionExtensions
+    {
+        public static IServiceCollection AddMemoryResponseCache(this IServiceCollection services)
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.AddOptions();
```
Tratcher: Aren't Options added by Hosting now?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/IResponseCache.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    public interface IResponseCache
+    {
+        object Get(string key);
+        void Set(string key, object entry);
```
JunTaoLuo: Yup, I added a TODO comment in `DistributedResponseCache` for it to track. Maybe I should move it here since it also affects IMemoryCache as well.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/MemoryResponseCache.cs
JunTaoLuo: These methods also throw exceptions, for example, if the underlying storage has been disposed.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/DistributedResponseCache.cs
JunTaoLuo: I'll move all the storage related types to the internal namespace and make them internal types since they shouldn't be needed outside of response caching.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/MemoryResponseCache.cs
Tratcher: That's not an error we should be hiding.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/CachedResponse.cs
```diff
@@ -3,12 +3,14 @@
 
 using Microsoft.AspNetCore.Http;
 
-namespace Microsoft.AspNetCore.ResponseCaching
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
```
Tratcher: your folder structure and namespaces should match

src/Microsoft.AspNetCore.ResponseCaching/ResponseCache/DefaultResponseCacheEntrySerializer.cs
```diff
@@ -0,0 +1,153 @@
...
+            
+            if (entry is CachedResponse)
+            {
+                WriteCachedResponse(writer, entry as CachedResponse);
+            }
+            else if (entry is CachedVaryBy)
+            {
+                WriteCachedVaryBy(writer, entry as CachedVaryBy);
+            }
+        }
```
Tratcher: else throw NotImplemented or NotSupported

Tratcher:
:shipit: after one comment

PR #15, opened 2016-08-05
Implement Http cache

JunTaoLuo:
todos:
- [x] Triage items marked with TODO: comment
- [x] Add more tests

cc @Tratcher @davidfowl


src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCache.cs
```diff
@@ -45,11 +45,17 @@ public void Remove(string key)
             }
         }
 
-        public void Set(string key, object entry)
+        public void Set(string key, object entry, TimeSpan validUntil)
```
Tratcher: `validUntil` is a strange name for a TimeSpan

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -107,10 +151,151 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+            // TODO: RFC also lists 203, 204, 206, 300, 301, 404, 405, 410, 414, and 501 as cacheable by default
+            if (HttpContext.Response.StatusCode != StatusCodes.Status200OK)
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        private bool ValidateEncoding(ResponseHeaders cachedResponsHeaders)
```
Tratcher: Isn't encoding a vary-by?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -107,10 +151,151 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+        {
+            // TODO: Content negotiation if there are multiple cached response formats?
+            return true;
+        }
+
+        private bool EntryIsFresh(ResponseHeaders responseHeaders, TimeSpan age, bool verifyAgainstRequest)
+        {
+            var responseCacheControl = responseHeaders.CacheControl ?? new CacheControlHeaderValue();
```
Tratcher: Keep a static Emtpy CacheControlHeaderValue?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,62 +306,76 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+                        foreach (var header in cachedResponse.Headers)
+                        {
+                            response.Headers.Add(header);
+                        }
+
+                        response.Headers[HeaderNames.Age] = ((int)Math.Min(age.TotalSeconds, int.MaxValue)).ToString(CultureInfo.InvariantCulture);
+
+                        if (_responseType == ResponseType.HeadersOnly)
+                        {
+                            response.ContentLength = 0;
```
Tratcher: Don't set Content-Length = 0 for HEAD requests, they should get the real content-length.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -186,34 +385,71 @@ internal void FinalizeCaching()
...
             }
 
             // TODO: TEMP, flush the buffer to the client
             Buffer.Seek(0, SeekOrigin.Begin);
             Buffer.CopyTo(OriginalResponseStream);
         }
 
+        internal void HookResponseStream()
+        {
+            // TODO: Use a wrapper stream to listen for writes (e.g. the start of the response),
```
Tratcher: Prioritize this please. Buffering all responses is a blocking issue.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: This is a little tricky. Technically yes, that's something that you can vary by, but generally this field varies significantly between different requests, even if they can be satisfied by the same encoding. I see in other implementations this field is treated differently so look for an encoding in the cache that satisfies the request. I'm looking into replicating this logic right now.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
Tratcher: missed one?

Tratcher:
So far so good.

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCache.cs
JunTaoLuo: `validFor`?

Tratcher: yeah, or expireAfter

test/Microsoft.AspNetCore.ResponseCaching.Tests/ResponseCachingTests.cs
```diff
@@ -16,35 +17,184 @@ public class ResponseCachingTests
...
+                headers.Date = DateTimeOffset.UtcNow;
+                headers.Headers["X-Value"] = uniqueId;
+                await context.Response.WriteAsync(uniqueId);
+            });
+
+            using (var server = new TestServer(builder))
+            {
+                var client = server.CreateClient();
+                var initialResponse = await client.GetAsync("");
+                client.DefaultRequestHeaders.CacheControl = new System.Net.Http.Headers.CacheControlHeaderValue()
```
JunTaoLuo: @Tratcher is there any reason we have our own version of `CacheControlHeaderValue`? It looks very similar to the one in `System.Net.Http.Headers`.

Tratcher: Yeah, we cloned most of System.Net.Http.Headers in order to make out-of-band usability and perf improvements.

src/Microsoft.AspNetCore.ResponseCaching/IResponseCache.cs
```diff
@@ -1,13 +1,14 @@
...
+using System;
+
 namespace Microsoft.AspNetCore.ResponseCaching
 {
     public interface IResponseCache
     {
         object Get(string key);
-        // TODO: Set expiry policy in the underlying cache?
-        void Set(string key, object entry);
+        void Set(string key, object entry, TimeSpan validFor);
```
davidfowl: We might need a context object here. What other things will we need to pass?

src/Microsoft.AspNetCore.ResponseCaching/Internal/Helpers.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
+{
+    internal static class Helpers
+    {
+        internal static Task CompletedTask { get; } = Task.FromResult(0);
+
+        internal static ConfiguredTaskAwaitable SupressContext(this Task task)
```
davidfowl: Not sure about these helpers. They kinda obscure the default API. Only WebListener uses these. We don't do it anywhere else in the stack.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -107,10 +162,165 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+            {
+                if (RequestCacheControl.NoCache || RequestCacheControl.NoStore)
+                {
+                    return false;
+                }
+            }
+            else
+            {
+                // Support for legacy HTTP 1.0 cache directive
+                var pragmaHeaderValues = request.Headers[HeaderNames.Pragma];
```
davidfowl: Does this only matter if we're looking a http 1.0 requests?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -107,10 +162,165 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+            var response = HttpContext.Response;
+
+            // Do not cache responses varying by *
+            if (string.Equals(response.Headers[HeaderNames.Vary], "*", StringComparison.OrdinalIgnoreCase))
+            {
+                return false;
+            }
+
+            // Do not cache responses that may bypass the response body
+            if (HttpContext.Features.Get<IHttpSendFileFeature>() != null)
```
davidfowl: This is an interesting one. It means middleware that may provide send file as a shim won't get responses cached.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -107,10 +162,165 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+        }
+
+        internal bool EntryIsFresh(ResponseHeaders responseHeaders, TimeSpan age, bool verifyAgainstRequest)
+        {
+            var responseCacheControl = responseHeaders.CacheControl ?? EmptyCacheControl;
+            
+            // Add min-fresh requirements
+            if (verifyAgainstRequest)
+            {
+                age += RequestCacheControl.MinFresh ?? TimeSpan.Zero;
```
davidfowl: Is MinFresh nullable?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
Tratcher: You could take this one step further and wrap the IHttpSendFileFeature if it's present and disable caching if it's called.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: Yes

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: More or less. http 1.1 requests might have them for back-compat reasons and when there is no cache control headers, it should be checked instead.

src/Microsoft.AspNetCore.ResponseCaching/Internal/Helpers.cs
JunTaoLuo: I do use them here: https://github.com/aspnet/ResponseCaching/pull/15/files#diff-90878ba706bfa37e215c2459d905ebe0R364

src/Microsoft.AspNetCore.ResponseCaching/IResponseCache.cs
JunTaoLuo: What do we need in the context object? Should we add that when we actually need it?

JunTaoLuo:
ping @davidfowl @Tratcher

src/Microsoft.AspNetCore.ResponseCaching/Internal/Helpers.cs
davidfowl: You shouldn't.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,99 +338,160 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+
+                        // Add a content-length if required
+                        if (response.ContentLength == null && string.IsNullOrEmpty(response.Headers[HeaderNames.TransferEncoding]))
+                        {
+                            response.ContentLength = body.Length;
+                        }
+
+                        if (body.Length > 0)
+                        {
+                            await response.Body.WriteAsync(body, 0, body.Length).SupressContext();
```
davidfowl: get rid of `SuppresContext()`.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -32,29 +34,35 @@ public ResponseCachingMiddleware(RequestDelegate next, IResponseCache cache)
...
-
                 try
                 {
+                    // Hook up to listen to the response stream
+                    cachingContext.HookResponseStream();
+
+                    // Subscribe to OnStarting event
+                    context.Response.OnStarting(state =>
+                    {
+                        (state as ResponseCachingContext).OnResponseStarting();
```
davidfowl: Don't as cast. Just use a normal cast

src/Microsoft.AspNetCore.ResponseCaching/project.json
```diff
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "Microsoft.AspNetCore.Http": "1.1.0-*",
+    "Microsoft.AspNetCore.Http.Extensions": "1.1.0-*",
```
davidfowl: Which methods do you use from here?

src/Microsoft.AspNetCore.ResponseCaching/Internal/DefaultResponseCacheEntrySerializer.cs
```diff
@@ -78,6 +81,7 @@ public static object Read(BinaryReader reader)
         // Body (byte[])
         private static CachedResponse ReadCachedResponse(BinaryReader reader)
         {
+            var created = DateTimeOffset.ParseExact(reader.ReadString(), RoundTripFormat, CultureInfo.InvariantCulture);
```
Tratcher: Storing UtcTicks might be more efficient / performant.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -32,29 +34,35 @@ public ResponseCachingMiddleware(RequestDelegate next, IResponseCache cache)
...
-                // Hook up to listen to the response stream
-                cachingContext.HookResponseStream();
-
                 try
                 {
+                    // Hook up to listen to the response stream
+                    cachingContext.HookResponseStream();
+
+                    // Subscribe to OnStarting event
+                    context.Response.OnStarting(state =>
```
Tratcher: store this lamda as a static delegate, otherwise it will create a new instance for every request.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -31,40 +45,87 @@ public ResponseCachingContext(HttpContext httpContext, IResponseCache cache)
...
+        internal bool CacheResponse
+        {
+            get
+            {
+                if (_cacheResponse == null)
+                {
+                    // TODO: apparent age vs corrected age value
+                    var responseAge = _responseTime - ResponseHeaders.Date ?? TimeSpan.Zero;
+
+                    _cacheResponse = ResponseStarted && ResponseIsCacheable() && EntryIsFresh(ResponseHeaders, responseAge, verifyAgainstRequest: false);
```
Tratcher: When will CacheResponse get called if ResponseStarted has not been set?

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheStream.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            await BufferedStream.WriteAsync(buffer, offset, count, cancellationToken);
+            await _innerStream.WriteAsync(buffer, offset, count, cancellationToken);
+        }
+
+        public override void WriteByte(byte value)
+        {
+            BufferedStream.WriteByte(value);
```
Tratcher: This is still fully buffering every response. When OnResponseStarting fires and it calculates CacheResponse, if it evaluates to false it should tell ResponseCacheStream to stop buffering and discard any already buffered data.

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheStream.cs
```diff
@@ -0,0 +1,75 @@
...
+        public override int Read(byte[] buffer, int offset, int count)
+            => _innerStream.Read(buffer, offset, count);
+
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            BufferedStream.Write(buffer, offset, count);
+            _innerStream.Write(buffer, offset, count);
+        }
+
+        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
```
Tratcher: You must override Begin/End Write as well, #if net451.

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheStream.cs
Tratcher: Actually, call the inner write first. If that does not trigger the OnStarting event and evaluate CacheResponse immediately, turn off caching for this response. This could happen if a buffering middlware was in the pipeline ahead of you, you want to avoid double-buffering the response.

Tratcher: And if the inner write ever throws disable caching and clear the buffer.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
Tratcher: File a bug for this, this middleware will never work on WebListener otherwise.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,99 +338,160 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
-                // TODO: Verify content freshness, or else re-validate the data?
-
                 var cachedResponse = cacheEntry as CachedResponse;
+                var cachedResponseHeaders = new ResponseHeaders(cachedResponse.Headers);
 
-                var response = HttpContext.Response;
-                // Copy the cached status code and response headers
-                response.StatusCode = cachedResponse.StatusCode;
-                foreach (var pair in cachedResponse.Headers)
+                _responseTime = DateTimeOffset.UtcNow;
```
Tratcher: Abstract out UtcNow with ISystemClock for testing?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,99 +338,160 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+                    // Copy the cached status code and response headers
+                    response.StatusCode = cachedResponse.StatusCode;
+                    foreach (var header in cachedResponse.Headers)
+                    {
+                        response.Headers.Add(header);
+                    }
 
-                // TODO: Update cache headers (Age)
-                response.Headers["Served_From_Cache"] = DateTime.Now.ToString();
+                    response.Headers[HeaderNames.Age] = ((int)Math.Min(age.TotalSeconds, int.MaxValue)).ToString(CultureInfo.InvariantCulture);
```
Tratcher: Why constrain it to int? what's wrong with long?

src/Microsoft.AspNetCore.ResponseCaching/project.json
JunTaoLuo: Typed headers.

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheStream.cs
Tratcher: There also needs to be a configurable size limit for what we're willing to cache.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: Ah, I forgot to update this. It used to be evaluated in several places but now it's only evaluated after OnResponseStarting so ResponseStarted will always be set to true. No need to check for it anymore

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,99 +338,160 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+                    if (header.Key != HeaderNames.Age
+                        && header.Key != HeaderNames.SetCookie)
+                    {
+                        _cachedResponse.Headers.Add(header);
+                    }
                 }
+            }
+            else
+            {
+                UnhookResponseStream();
```
Tratcher: This should only be called during middleware unwind, not in OnStarting

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,99 +338,160 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+                    Created = ResponseHeaders.Date.Value,
+                    StatusCode = HttpContext.Response.StatusCode
                 };
-                foreach (var pair in HttpContext.Response.Headers)
+
+                foreach (var header in ResponseHeaders.Headers)
                 {
-                    cachedResponse.Headers[pair.Key] = pair.Value;
+                    if (header.Key != HeaderNames.Age
+                        && header.Key != HeaderNames.SetCookie)
```
Tratcher: Case insensitive header key compairisons

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: Why not do it earlier if we realize that we won't be caching anyways?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -12,6 +12,11 @@ public class ResponseCachingMiddleware
     {
         private readonly RequestDelegate _next;
         private readonly IResponseCache _cache;
+        private static readonly Func<object, Task> OnStartingCallback = state =>
```
Tratcher: nit: static fields should be first.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -32,29 +37,31 @@ public ResponseCachingMiddleware(RequestDelegate next, IResponseCache cache)
...
                     return;
                 }
 
-                // Hook up to listen to the response stream
-                cachingContext.HookResponseStream();
-
                 try
                 {
+                    // Hook up to listen to the response stream
+                    cachingContext.HookResponseStream();
```
Tratcher: why was this moved inside the try/finally? If it fails throws what will UnhookResponseStream do?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: We could. RFC says minimum of 31 bits of non-negative integer range so `int` is enough but that's a minimum requirement so I guess we can also use long.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -32,29 +37,31 @@ public ResponseCachingMiddleware(RequestDelegate next, IResponseCache cache)
...
-                    {
-                        cachingContext.OnResponseStarting();
-                    }
+                    cachingContext.OnResponseStarting();
+
                     // Finalize the cache entry
-                    cachingContext.FinalizeCaching();
+                    cachingContext.FinalizeCachingBody();
                 }
                 finally
```
Tratcher: catch { disable response caching; throw; }

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
Tratcher: OnStarting is only for setting headers, you don't have full control of the response. This could break upstream wrappers.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
JunTaoLuo: Strange I missed that, must be a bad refactoring.

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheStream.cs
JunTaoLuo: Was depending on unhooking the response stream instead of disabling the cache but that won't work. Will file an issue for a size limit.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
JunTaoLuo: No need, response caching is only completed if everything succeeds. Early exit by default behaves as if caching was disabled.

src/Microsoft.AspNetCore.ResponseCaching/Internal/SendFileFeatureWrapper.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            _originalSendFileFeature = originalSendFileFeature;
+            _responseCacheStream = responseCacheStream;
+        }
+
+        // Flush and disable the buffer if anyone tries to call the SendFile feature.
+        public async Task SendFileAsync(string path, long offset, long? length, CancellationToken cancellation)
+        {
+            _responseCacheStream.DisableBuffering();
+            await _originalSendFileFeature.SendFileAsync(path, offset, length, cancellation);
```
Tratcher: no async/await required.

test/Microsoft.AspNetCore.ResponseCaching.Tests/ResponseCachingContextTests.cs
```diff
@@ -190,20 +190,6 @@ public void ResponseIsCacheable_VaryByStar_NotAllowed()
         }
 
         [Fact]
-        public void ResponseIsCacheable_IHttpSendFileFeature_NotAllowed()
```
Tratcher: Test for disabling cache instead?

Tratcher: Nevermind

src/Microsoft.AspNetCore.ResponseCaching/Internal/DefaultResponseCacheEntrySerializer.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.AspNetCore.ResponseCaching.Internal
     internal static class DefaultResponseCacheSerializer
     {
         private const int FormatVersion = 1;
+        private const string RoundTripFormat = "o";
```
Tratcher: No longer used

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheStream.cs
```diff
@@ -0,0 +1,184 @@
...
+            {
+                _innerStream.SetLength(value);
+                if (BufferingEnabled)
+                {
+                    BufferedStream.SetLength(value);
+                }
+            }
+            catch
+            {
+                DisableBuffering();
```
Tratcher: ?

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheStream.cs
```diff
@@ -0,0 +1,184 @@
...
+                throw;
+            }
+        }
+
+        public override void Flush()
+            => _innerStream.Flush();
+
+        public override async Task FlushAsync(CancellationToken cancellationToken)
+            => await _innerStream.FlushAsync();
```
Tratcher: no async/await needed.

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheStream.cs
```diff
@@ -0,0 +1,184 @@
...
+        {
+            try
+            {
+                _innerStream.Write(buffer, offset, count);
+                if (BufferingEnabled)
+                {
+                    BufferedStream.Write(buffer, offset, count);
+                }
+            }
+            catch
```
Tratcher: These try/catch's only need to be around the _innerStream operations.

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheStream.cs
JunTaoLuo: `SetLength` can throw for the underlying stream?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,99 +346,173 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+                            response.ContentLength = body.Length;
+                        }
+
+                        if (body.Length > 0)
+                        {
+                            await response.Body.WriteAsync(body, 0, body.Length);
+                        }
+
+                        responseServed = true;
+                    }
```
Tratcher: else throw NotImplmenetedException(responseType)

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,99 +346,173 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+                    // TODO: Validate with endpoint instead
                 }
-                return true;
             }
 
-            return false;
-        }
+            if (!responseServed && RequestCacheControl.OnlyIfCached)
+            {
+                HttpContext.Response.Clear();
```
Tratcher: Why do you need the Clear()? It looks like you only set headers in responseServed scenarios.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,99 +346,173 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+        }
 
-                Cache.Set(_cacheKey, cachedResponse); // TODO: Timeouts
+        internal void FinalizeCachingBody()
+        {
+            if (CacheResponse && ResponseCacheStream.BufferingEnabled)
+            {
+                _cachedResponse.Body = ResponseCacheStream.BufferedStream.ToArray();
+
+                // TODO: Timeouts
```
Tratcher: You've fulfilled this TODO, or do you have more plans?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,99 +346,173 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
-        internal void FinalizeCaching()
+        internal void FinalizeCachingHeaders()
         {
             if (CacheResponse)
             {
+                // Create the cache entry now
                 var response = HttpContext.Response;
-                var varyHeaderValue = response.Headers["Vary"];
+                var varyHeaderValue = response.Headers[HeaderNames.Vary];
+                _cachedResponseValidFor = ResponseCacheControl.SharedMaxAge
```
Tratcher: This should have a configurable max value. It doesn't need to be huge because the caches have an automatic sliding refresh.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: I think it's sufficient to silently not serve a response when a response type is not specified. Even if we do throw an exception here, I think `InvalidOperationException` is more appropriate since the `_responseType` was `null`. The message would say something like "Response type must be set. RequestIsCacheable() must be called before trying to serve response from cache." Might also want to extract out the part inside `RequestIsCacheable` that sets this value in that case.

Tratcher: A missing response type here is a code bug that should throw in testing.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: Will probably add this in another PR where I add some configurability to the middleware along with other settings.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: Needed this to clear headers if we didn't throw for unsuccessful serving due to wrong response types. But will remove now.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -121,99 +346,175 @@ private string CreateVaryByCacheKey(CachedVaryBy varyBy)
...
+                        if (body.Length > 0)
+                        {
+                            await response.Body.WriteAsync(body, 0, body.Length);
+                        }
+
+                        responseServed = true;
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException("Response type not specified or is unrecognized.");
```
Tratcher: Include the value in the message.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -46,23 +54,26 @@ public ResponseCachingMiddleware(RequestDelegate next, IResponseCache cache)
...
                 {
                     cachingContext.UnhookResponseStream();
+
```
Tratcher: empty line?

Tratcher:
:shipit:

PR #16, opened 2016-08-23
Add configurable options for ResponseCaching

JunTaoLuo:
Override RequestIsCacheable
Override ResponseIsCacheable
Append customized cache key

Will add more configurability options as features are added.


src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        /// <summary>
+        /// Override default behavior for determining cacheability of an HTTP response.
+        /// </summary>
+        public Func<HttpContext, bool> ResponseIsCacheableOverride { get; set; }
+
+        /// <summary>
+        /// Create a custom segment that is appended to the default cache key.
+        /// </summary>
+        public Func<HttpContext, string> CreateCustomizedCacheKey { get; set; }
```
JunTaoLuo: We can consider passing the StringBuilder that contains the default key to the user. It may be more efficient if the user needs to concatenate many small segments since it saves creating a new instance of StringBuilder. But not sure how often this would arise and it may be a premature optimization. It would also have the side effect of the user being able to overwrite the default cache key.

Tratcher: They can get a string builder from the request services caching service

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -29,21 +30,25 @@ public class ResponseCachingContext
...
-        public ResponseCachingContext(HttpContext httpContext, IResponseCache cache)
-            : this(httpContext, cache, new SystemClock())
+        public ResponseCachingContext(HttpContext httpContext, IResponseCache cache, ResponseCachingOptions options)
+            : this(httpContext, cache, options, new SystemClock())
         {
         }
 
         // Internal for testing
-        internal ResponseCachingContext(HttpContext httpContext, IResponseCache cache, ISystemClock clock)
+        internal ResponseCachingContext(HttpContext httpContext, IResponseCache cache, ResponseCachingOptions options, ISystemClock clock)
```
Tratcher: Everywhere else we have the ISystemClock on the Options so you don't need this special constructor.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -145,46 +153,48 @@ internal string CreateCacheKey(CachedVaryBy varyBy)
...
-                    .Append(";");
             }
+            // TODO: Parse querystring params
 
-            // Parse querystring params
+            // Append custom cache key segment
+            if (Options.CreateCustomizedCacheKey != null)
+            {
+                builder.Append(";")
+                    .Append(Options.CreateCustomizedCacheKey(HttpContext));
```
Tratcher: Interesting, they can only append to the key, not create their own from scratch?

Tratcher: That's not obvious from the event name

JunTaoLuo: That's what we wanted? I specified that in the xml doc of the options.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -145,46 +153,48 @@ internal string CreateCacheKey(CachedVaryBy varyBy)
...
 
             return builder.ToString();
         }
 
         internal bool RequestIsCacheable()
         {
+            // Use optional override if specified by user
+            if (Options.RequestIsCacheableOverride != null)
+            {
+                return Options.RequestIsCacheableOverride(HttpContext);
```
Tratcher: This needs to return a try-state result. 1) Yes, 2) No, and 3) use the normal logic

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -236,6 +246,12 @@ internal bool RequestIsCacheable()
 
         internal bool ResponseIsCacheable()
         {
+            // Use optional override if specified by user
+            if (Options.ResponseIsCacheableOverride != null)
+            {
+                return Options.ResponseIsCacheableOverride(HttpContext);
```
Tratcher: Same as RequestIsCacheableOverride

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingOptions.cs
Tratcher: Also, interfaces prevent adding events. See https://github.com/aspnet/Security/issues/942

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: Hmm... should we swallow user exceptions like that? Why not just let it fail the request?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
Tratcher: The name needs work, they can't customize the whole key.

JunTaoLuo: I guess the naming could be better but something like `CreateCacheKeySegmentToBeAppended` sounds awkward too. `CreateCustomCacheKeyPostfix`?

Tratcher: CreateCustomKeySuffix?

Tratcher: Why would you use it? As a cuztomizable vary-by?

JunTaoLuo: Yup

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -145,46 +153,48 @@ internal string CreateCacheKey(CachedVaryBy varyBy)
...
+                    var value = HttpContext.Request.Headers[header];
 
-            foreach (var header in varyBy.Headers)
-            {
-                // TODO: Normalization of order, case?
-                var value = HttpContext.Request.Headers[header].ToString();
+                    // TODO: How to handle null/empty string?
+                    if (StringValues.IsNullOrEmpty(value))
+                    {
+                        value = "null";
```
muratg: Why?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -145,46 +153,48 @@ internal string CreateCacheKey(CachedVaryBy varyBy)
...
-        {
-            // TODO: resolve key format and delimiters
-            if (varyBy == null || varyBy.Headers.Count == 0)
+            if (varyBy?.Headers.Count > 0)
             {
-                return string.Empty;
-            }
-
-            var builder = new StringBuilder(";");
+                // TODO: resolve key format and delimiters
```
muratg: Before RTM?

JunTaoLuo: That's the plan.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: If the user explicitly specified to vary by a header, we use a placeholder value of `null` to indicate a missing value. This differentiates between an empty value of a request with an explicit varyby verses a request that does not varyby that header. May revisit this decision which is why I added a TODO item here. Maybe it makes sense to make this behaviour configurable.

src/Microsoft.AspNetCore.ResponseCaching/OverrideResult.cs
```diff
@@ -5,8 +5,8 @@ namespace Microsoft.AspNetCore.ResponseCaching
 {
     public enum OverrideResult
     {
-        True,
+        Ignore,
```
Tratcher: UseDefaultLogic

src/Microsoft.AspNetCore.ResponseCaching/OverrideResult.cs
```diff
@@ -5,8 +5,8 @@ namespace Microsoft.AspNetCore.ResponseCaching
...
         False,
```
Tratcher: NoCache

src/Microsoft.AspNetCore.ResponseCaching/OverrideResult.cs
```diff
@@ -5,8 +5,8 @@ namespace Microsoft.AspNetCore.ResponseCaching
...
-        Ignore
+        True
```
Tratcher: Cache

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingOptions.cs
```diff
@@ -1,31 +1,13 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.ResponseCaching;
 using Microsoft.AspNetCore.ResponseCaching.Internal;
 
 namespace Microsoft.AspNetCore.Builder
 {
     public class ResponseCachingOptions
```
JunTaoLuo: Guess this isn't needed anymore

src/Microsoft.AspNetCore.ResponseCaching/OverrideResult.cs
JunTaoLuo: DoNotCache?

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCachingResponseCacheabilityValidator.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    public interface IResponseCachingResponseCacheabilityValidator
```
davidfowl: Maybe merge this with the response one. Is the logic usually coupled or is it completely separate?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingServiceCollectionExtensions.cs
```diff
@@ -31,9 +32,19 @@ public static IServiceCollection AddDistributedResponseCache(this IServiceCollec
             }
             
             services.AddDistributedMemoryCache();
+            services.AddNoopResponseCachingServices();
             services.TryAdd(ServiceDescriptor.Singleton<IResponseCache, DistributedResponseCache>());
 
             return services;
         }
+
+        private static IServiceCollection AddNoopResponseCachingServices(this IServiceCollection services)
```
davidfowl: Get rid of the Noop name.

src/Microsoft.AspNetCore.ResponseCaching/Internal/NoopResponseCachingCacheKeySuffixProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
+{
+    public class NoopResponseCachingCacheKeySuffixProvider : IResponseCachingCacheKeySuffixProvider
+    {
+        public string CreateCustomKeySuffix(HttpContext httpContext) => null;
```
davidfowl: Why not implement the default behavior in here?

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCachingResponseCacheabilityValidator.cs
JunTaoLuo: I thought about that but the problem is that if you only need to override one of the cacheability checks, you still need to implement both. The argument can also be made from the other side where if you use need to use both overrides, now you need two different implementations. So we just need to pick our poison.

src/Microsoft.AspNetCore.ResponseCaching/Internal/NoopResponseCachingCacheKeySuffixProvider.cs
JunTaoLuo: I think during one of our meetings (or maybe just a offline discussion) we decided that we wanted to always keep the default key and only allow users to append their custom strings as a customized varyby. If we implement the default behaviour here, we give the user complete control over the cache key, is that what we want?

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCachingRequestCacheabilityValidator.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    public interface IResponseCachingRequestCacheabilityValidator
```
JunTaoLuo: These interface names are long and confusing...

JunTaoLuo:
:up::date:

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCachingResponseCacheabilityValidator.cs
Tratcher: Combine them, ` => OverrideResult.UseDefaultLogic;` is easy enough.

JunTaoLuo:
ping

Tratcher:
:shipit:

PR #17, opened 2016-08-24
Update no-store behaviour

JunTaoLuo:
I didn't follow the RFC regarding `no-store` directive in the request. This simply means the corresponding response cannot be cached. However, if a response exist in the cache already, it can be served.

cc @Tratcher @davidfowl


JunTaoLuo:
ping @Tratcher @davidfowl

Tratcher:
:shipit:

PR #20, opened 2016-08-26
Pool StringBuilder

JunTaoLuo:
So we don't create one per request.

cc @Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
@@ -10,14 +10,23 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Http.Headers;
 using Microsoft.AspNetCore.ResponseCaching.Internal;
+using Microsoft.Extensions.ObjectPool;
 using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.ResponseCaching
 {
-    public class ResponseCachingContext
+    internal class ResponseCachingContext
     {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -143,49 +145,61 @@ internal string CreateCacheKey()
 
         internal string CreateCacheKey(CachedVaryBy varyBy)
         {
-            var request = HttpContext.Request;
-            var builder = new StringBuilder()
-                .Append(request.Method.ToUpperInvariant())
-                .Append(";")
-                .Append(request.Path.Value.ToUpperInvariant());
+            var request = _httpContext.Request;
+            var builder = _builderPool?.Get() ?? new StringBuilder();
```
JunTaoLuo: One less visible caveat is that the default StringBuilder pool will has a max capacity of 4*1024 characters. If the generated key is larger than that value, no pooling will occur and we are back to allocating one per request. However, The user can replace the default policy to increase the max capacity. Also, 4*1024 characters is quite a large value so I'm not sure how likely this limit is hit.

Tratcher: Why can _builderPool be null?

JunTaoLuo: Can't really be null in production. I set it to null in the tests though. https://github.com/aspnet/ResponseCaching/pull/20/files/3ea31873704d140183cb4f09bcd1b615f110591b#diff-ade4e3024ab6def5a1a9db255f083dd4R659

Tratcher: yeah

JunTaoLuo:
:up::date:

Tratcher:
:shipit

PR #21, opened 2016-08-26
Add support for vary by query string params

JunTaoLuo:
cc @Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/Internal/DefaultResponseCacheEntrySerializer.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/Internal/DefaultResponseCacheEntrySerializer.cs
@@ -96,12 +96,26 @@ namespace Microsoft.AspNetCore.ResponseCaching.Internal
         }
 
         // Serialization Format
-        // Headers (comma separated string)
+        // Headers count 
+        // Headers if count > 0 (comma separated string)
+        // Params count 
+        // Params if count > 0 (comma separated string)
         private static CachedVaryBy ReadCachedVaryBy(BinaryReader reader)
         {
-            var headers = reader.ReadString().Split(',');
+            var headerCount = reader.ReadInt32();
+            var headers = new string[headerCount];
+            for (var index = 0; index < headerCount; index++)
+            {
+                headers[index] = reader.ReadString();
+            }
+            var paramCount = reader.ReadInt32();
+            var param = new string[paramCount];
+            for (var index = 0; index < paramCount; index++)
+            {
+                param[index] = reader.ReadString();
+            }
 
-            return new CachedVaryBy { Headers = headers };
+            return new CachedVaryBy { Headers = headers, Params = param };
         }
 
         // See serialization format above
@@ -154,7 +168,18 @@ namespace Microsoft.AspNetCore.ResponseCaching.Internal
         private static void WriteCachedVaryBy(BinaryWriter writer, CachedVaryBy entry)
         {
             writer.Write(nameof(CachedVaryBy));
-            writer.Write(entry.Headers);
+
+            writer.Write(entry.Headers.Count);
+            foreach (var header in entry.Headers)
+            {
+                writer.Write(header);
+            }
+
+            writer.Write(entry.Params.Count);
+            foreach (var param in entry.Params)
+            {
+                writer.Write(param);
+            }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
@@ -4,6 +4,7 @@
 using System;
 using System.IO;
 using System.Globalization;
+using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
@@ -19,6 +20,8 @@ namespace Microsoft.AspNetCore.ResponseCaching
     internal class ResponseCachingContext
     {
         private static readonly CacheControlHeaderValue EmptyCacheControl = new CacheControlHeaderValue();
+        // Use the record separator for delimiting components of the cache key to avoid possible collisions
+        private static readonly char KeyDelimiter = '\x1e';
 
         private readonly HttpContext _httpContext;
         private readonly IResponseCache _cache;
@@ -150,13 +153,19 @@ namespace Microsoft.AspNetCore.ResponseCaching
 
             try
             {
+                // Default key 
                 builder
                     .Append(request.Method.ToUpperInvariant())
-                    .Append(";")
+                    .Append(KeyDelimiter)
                     .Append(request.Path.Value.ToUpperInvariant());
 
+                // Vary by headers
                 if (varyBy?.Headers.Count > 0)
                 {
+                    // Append a group separator for the header segment of the cache key
+                    builder.Append(KeyDelimiter)
+                        .Append('H');
+
                     // TODO: resolve key format and delimiters
                     foreach (var header in varyBy.Headers)
                     {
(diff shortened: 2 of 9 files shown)


src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -175,7 +178,42 @@ internal string CreateCacheKey(CachedVaryBy varyBy)
...
                 }
-                // TODO: Parse querystring params
+
+                // Vary by query params
+                if (varyBy?.Params.Count > 0)
+                {
+                    if (varyBy.Params.Count == 1 && string.Equals(varyBy.Params[0], "*", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // Vary by all available query params
+                        foreach (var query in _httpContext.Request.Query.OrderBy(q => q.Key))
```
JunTaoLuo: A few issues here. 
1) order by case sensitivity? The QueryCollection operates on case-insensitive keys so maybe we need to order by `IgnoreCase` and convert all query keys to upper case. Sign perf...
2) Linq vs Array.Sort(). Was using Linq to guarantee a stable sort but I think the keys are unique so it doesn't matter? Also, what is the impact on performance?
3) Could consider using the raw `QueryString` but that approach will forgo order-insensitivity of the params. I don't think we should do this unless perf is a big problem here.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -169,19 +177,62 @@ internal string CreateCacheKey(CachedVaryBy varyBy)
...
+                if (varyBy?.Params.Count > 0)
+                {
+                    // Append a value denoting the start of the query parameter segment of the cache key
+                    builder.Append(KeyDelimiter)
+                        .Append('Q');
+
+                    if (varyBy.Params.Count == 1 && string.Equals(varyBy.Params[0], "*"))
+                    {
+                        // Vary by all available query params
+                        foreach (var query in _httpContext.Request.Query.OrderBy(q => q.Key))
```
JunTaoLuo: Strange this comment got deleted, reposting.
A few issues here. 
1) order by case sensitivity? The QueryCollection operates on case-insensitive keys so maybe we need to order by IgnoreCase and convert all query keys to upper/lower case. Sigh perf...
2) Linq vs Array.Sort() or List.Sort(). Was using Linq to guarantee a stable sort but I think the keys are unique so it doesn't matter. Need to check the impact on performance.
3) Could consider using the raw QueryString but that approach will forgo order-insensitivity of the params. I don't think we should do this unless perf is a big problem here.

src/Microsoft.AspNetCore.ResponseCaching/Internal/DefaultResponseCacheEntrySerializer.cs
```diff
@@ -154,7 +160,18 @@ private static void WriteCachedResponse(BinaryWriter writer, CachedResponse entr
         private static void WriteCachedVaryBy(BinaryWriter writer, CachedVaryBy entry)
         {
             writer.Write(nameof(CachedVaryBy));
-            writer.Write(entry.Headers);
+
+            writer.Write(entry.Headers.Count);
+            if (entry.Headers.Count > 0)
+            {
+                writer.Write(entry.Headers);
```
Tratcher: You can save yourself some joins and splits by adding a for loop here and writing out the headers and param values separately.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -19,6 +20,7 @@ namespace Microsoft.AspNetCore.ResponseCaching
     internal class ResponseCachingContext
     {
         private static readonly CacheControlHeaderValue EmptyCacheControl = new CacheControlHeaderValue();
+        private static readonly char KeyDelimiter = '\x1e'; // Use the record separator for delimiting components of the cache key
```
Tratcher: ...to avoid possible collisions

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -150,13 +152,19 @@ internal string CreateCacheKey(CachedVaryBy varyBy)
...
                     .Append(request.Method.ToUpperInvariant())
-                    .Append(";")
+                    .Append(KeyDelimiter)
                     .Append(request.Path.Value.ToUpperInvariant());
 
+                // Vary by headers
                 if (varyBy?.Headers.Count > 0)
                 {
+                    // Append a value denoting the start of the header segment of the cache key
+                    builder.Append(KeyDelimiter)
```
Tratcher: How about "Group Separator" here?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -169,19 +177,62 @@ internal string CreateCacheKey(CachedVaryBy varyBy)
...
+                        .Append('Q');
+
+                    if (varyBy.Params.Count == 1 && string.Equals(varyBy.Params[0], "*"))
+                    {
+                        // Vary by all available query params
+                        foreach (var query in _httpContext.Request.Query.OrderBy(q => q.Key))
+                        {
+                            builder.Append(KeyDelimiter)
+                                .Append(query.Key.ToLower())
```
Tratcher: Always ToUpperInvariant

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -169,19 +177,62 @@ internal string CreateCacheKey(CachedVaryBy varyBy)
...
+                                .Append("=")
+                                .Append(query.Value);
```
Tratcher: They key is case insensitive but the value is case sensitive?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingHttpContextExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    // TODO: Temporary interface for endpoints to specify options for response caching
+    public static class ResponseCachingHttpContextExtensions
+    {
+        public static void AddResponseCachingOptions(this HttpContext httpContext)
+        {
+            httpContext.Features.Set(new ResponseCachingOptionsFeature());
```
Tratcher: ResponseCachingFeature, Options is something else

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingOptionsFeature.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    // TODO: Temporary interface for endpoints to specify options for response caching
+    public class ResponseCachingOptionsFeature
+    {
+        public StringValues Params;
```
Tratcher: { get; set; }

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: It's ambiguous what we need to do here, we'll discuss offline in a design meeting.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingFeature.cs
```diff
@@ -6,8 +6,8 @@
 namespace Microsoft.AspNetCore.ResponseCaching
 {
     // TODO: Temporary interface for endpoints to specify options for response caching
-    public class ResponseCachingOptionsFeature
+    public class ResponseCachingFeature
     {
-        public StringValues Params;
+        public StringValues Params { get; set; }
```
Tratcher: VaryByParams?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingHttpContextExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            httpContext.Features.Set(new ResponseCachingFeature());
+        }
+
+        public static void RemoveResponseCachingOptions(this HttpContext httpContext)
+        {
+            httpContext.Features.Set<ResponseCachingFeature>(null);
+        }
+
+        public static ResponseCachingFeature GetResponseCachingOptions(this HttpContext httpContext)
```
Tratcher: Feature

Tratcher:
:shipit:

PR #22, opened 2016-08-29
Conditional requests

JunTaoLuo:
Add support to send 304 responses for requests conditional on `IfModifiedSince` and `IfNoneMatch`.

cc @Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
@@ -443,41 +443,50 @@ namespace Microsoft.AspNetCore.ResponseCaching
 
                 if (EntryIsFresh(cachedResponseHeaders, age, verifyAgainstRequest: true))
                 {
-                    var response = _httpContext.Response;
-                    // Copy the cached status code and response headers
-                    response.StatusCode = cachedResponse.StatusCode;
-                    foreach (var header in cachedResponse.Headers)
+                    // Check conditional request rules
+                    if (ConditionalRequestSatisfied(cachedResponseHeaders))
                     {
-                        response.Headers.Add(header);
-                    }
-
-                    response.Headers[HeaderNames.Age] = age.TotalSeconds.ToString("F0", CultureInfo.InvariantCulture);
-
-                    if (_responseType == ResponseType.HeadersOnly)
-                    {
-                        responseServed = true;
-                    }
-                    else if (_responseType == ResponseType.FullReponse)
-                    {
-                        // Copy the cached response body
-                        var body = cachedResponse.Body;
-
-                        // Add a content-length if required
-                        if (response.ContentLength == null && string.IsNullOrEmpty(response.Headers[HeaderNames.TransferEncoding]))
-                        {
-                            response.ContentLength = body.Length;
-                        }
-
-                        if (body.Length > 0)
-                        {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -443,6 +443,13 @@ internal bool EntryIsFresh(ResponseHeaders responseHeaders, TimeSpan age, bool v
 
                 if (EntryIsFresh(cachedResponseHeaders, age, verifyAgainstRequest: true))
                 {
+                    // Check conditional request rules
+                    if (ConditionalRequestSatisfied(cachedResponseHeaders))
+                    {
+                        _httpContext.Response.StatusCode = StatusCodes.Status304NotModified;
+                        return true;
```
Tratcher: inconsistent use of `responseServed` vs `return true`.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -489,13 +496,41 @@ internal bool EntryIsFresh(ResponseHeaders responseHeaders, TimeSpan age, bool v
...
-                responseServed = true;
+                return true;
             }
 
             return responseServed;
         }
 
+        internal bool ConditionalRequestSatisfied(ResponseHeaders cachedResponseHeaders)
+        {
+            if (RequestHeaders.IfNoneMatch != null)
```
Tratcher: This property parses every time you access it, with no caching. Put it in a local variable

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: I had the `responseServed` variable previously because we couldn't short-circuit after failing to serve a response due to the `OnlyIfCached` header. The conditional requests are different, if we served that, we can return right away. But I guess for consistency, I can just keep using `responseServed` and return it at the end. This means one extra check on the boolean even after satisfying the conditional request but I guess that's fine.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
JunTaoLuo: Good catch

JunTaoLuo:
:up::date:

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -489,13 +498,43 @@ internal bool EntryIsFresh(ResponseHeaders responseHeaders, TimeSpan age, bool v
...
+                if (ifNoneMatchHeader.Count == 1 && ifNoneMatchHeader[0].Equals(EntityTagHeaderValue.Any))
+                {
+                    return true;
+                }
+
+                if (cachedResponseHeaders.ETag != null)
+                {
+                    foreach (var tag in ifNoneMatchHeader)
+                    {
+                        // TODO: use strong comparison
```
Tratcher: This TODO is complete

Tratcher:
:shipit:
What about all the other conditionals?

JunTaoLuo:
Only `IfNoneMatch` and `IfModifiedSince` applies to us for now.

PR #23, opened 2016-08-30
Update calls for string comparison

JunTaoLuo:
Updating or replacing calls for string comparison.

cc @Tratcher @troydai


Tratcher:
:shipit:

PR #24, opened 2016-09-01
Allow the prepending of a cache key prefix

JunTaoLuo:
cc @Tratcher @sebastienros

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCachingCacheKeyModifier.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCachingCacheKeyModifier.cs
@@ -5,13 +5,13 @@ using Microsoft.AspNetCore.Http;
 
 namespace Microsoft.AspNetCore.ResponseCaching
 {
-    public interface IResponseCachingCacheKeySuffixProvider
+    public interface IResponseCachingCacheKeyModifier
     {
         /// <summary>
-        /// Create a key segment that is appended to the default cache key.
+        /// Create a key segment that is prepended to the default cache key.
         /// </summary>
         /// <param name="httpContext">The <see cref="HttpContext"/>.</param>
-        /// <returns>The key segment that will be appended to the default cache key.</returns>
-        string CreateCustomKeySuffix(HttpContext httpContext);
+        /// <returns>The key segment that will be prepended to the default cache key.</returns>
+        string CreatKeyPrefix(HttpContext httpContext);
     }
 }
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCachingCacheKeyModifier.cs
```diff
@@ -5,13 +5,20 @@
...
 {
-    public interface IResponseCachingCacheKeySuffixProvider
+    public interface IResponseCachingCacheKeyModifier
     {
         /// <summary>
+        /// Create a key segment that is prepended to the default cache key.
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/>.</param>
+        /// <returns>The key segment that will be prepended to the default cache key.</returns>
+        string CreatKeyPrefix(HttpContext httpContext);
```
Tratcher: I don't see why we need both of these. The key is never re-parsed, so it doesn't matter to us if the customization is at the start or end. The request for a prefix was for shared cache disambiguation, correct? If that's the case then the suffix data could just as easily be put at the beginning too.

sebastienros: I personally care only about prefixing as this is the most obvious visual hint  on a list, and will probably be better supported than suffixes to filter items in the cache. But then why not just let the service change the key freely?

Otherwise, if it's to provide a simple string to append somewhere, an option might be sufficient.

JunTaoLuo: A few interactions we might need to consider here:
1) Hashing
Not sure if this is something we want to implement yet but if we do, do we want to keep the prefix unhashed for ease of identification and the suffix to be hashed along with the rest of the key?
2) Base key vs VaryBy key
This won't affect the user but do we the suffix to be in the base key or a part of the vary-by key?

I agree, functionally there's no reason to have both a prefix and a suffix other than aesthetics when inspecting the cache, but I thought that was the point?

We can also add a default service for creating keys and let the user override that.

sebastienros: My view is that we need at least a prefix, and the prefix is not hashed or part of the hash.
A simple configuration option at the middleware level would be fine. Even with multi-tenancy where each tenant would have its own middleware instance.

JunTaoLuo: The pattern we want to use is services instead of configuration options on the middleware to allow resolution of other services from DI.

JunTaoLuo:
:up::date: let's just a prefix for now. If there is a reason to add a suffix we'll add it when the time comes

Tratcher:
:shipit:

PR #25, opened 2016-09-01
Add an option to specify the max response size cached

JunTaoLuo:
cc @Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
@@ -7,6 +7,7 @@ using System.Globalization;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Http.Headers;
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+using Microsoft.AspNetCore.ResponseCaching.Internal;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public class ResponseCachingOptions
+    {
+        /// <summary>
+        /// The largest cacheable size for the response body in bytes.
+        /// </summary>
+        public int? MaximumCachedBodySize { get; set; }
```
JunTaoLuo: Should we specify a default size? Also could consider the naming `MaximumCacheableBodySize` or `MaximumBodySizeToCache`.

Tratcher: `long`, not `int?`. Default to 1mb for starters?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -598,7 +594,9 @@ internal void FinalizeCachingHeaders()
 
         internal void FinalizeCachingBody()
         {
-            if (CacheResponse && ResponseCacheStream.BufferingEnabled)
+            if (CacheResponse &&
+                ResponseCacheStream.BufferingEnabled &&
+                !(_options.MaximumCachedBodySize < ResponseCacheStream.BufferedStream.Length))
```
Tratcher: BufferingEnabled should be false as soon as MaximumCachedBodySize is exceeded

JunTaoLuo:
:up::date:

Tratcher:
FYI: It looks like MemoryStream.Dispose doesn't actually free the memory. You need to call SetLength and Capacity to shrink it.

Tratcher:
:shipit:

PR #26, opened 2016-09-02
Add options to configure case sensitivity of request paths

JunTaoLuo:
cc @Tratcher @davidfowl


src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingOptions.cs
```diff
@@ -14,6 +14,11 @@ public class ResponseCachingOptions
         public long MaximumCachedBodySize { get; set; } = 1024 * 1024;
 
         /// <summary>
+        /// <c>true</c> if request paths are case-sensitive; otherwise <c>false</c>.
```
BrennanConroy: Should the summary say the default is false?

JunTaoLuo:
:up::date:

Tratcher:
:shipit:

PR #28, opened 2016-09-06
Refactor ResponseCachingContext

JunTaoLuo:
- [x] extract cache key generation into a service
- [x] extract cacheability checks into a service
- [x] make responses with set-cookies uncacheable 

might also consider adding to this PR if difficult to implement separately:
- [ ] separate header and body
- [ ] chunk body
- [ ] GUID based keys


src/Microsoft.AspNetCore.ResponseCaching/DefaultCacheabilityValidator.cs
```diff
@@ -0,0 +1,214 @@
...
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Headers;
+using Microsoft.AspNetCore.ResponseCaching.Internal;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    public class DefaultCacheabilityValidator : IResponseCachingCacheabilityValidator
```
Tratcher: Remove `Default`

Tratcher: IResponseCaching is redundant with the namespace

JunTaoLuo: Could probably simplify the names of a bunch of other stuff as well probably.

JunTaoLuo:
The PR is getting a little large so let's not fuel the fire with more features and just deal with the extractions of `IKeyProvider` and `ICacheabilityValidator`.

cc @Tratcher @davidfowl

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingState.cs
```diff
@@ -0,0 +1,89 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Headers;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
+{
+    internal class ResponseCachingState
```
Tratcher: more public?

src/Microsoft.AspNetCore.ResponseCaching/Internal/StreamWrapper.cs
```diff
@@ -8,12 +8,12 @@
 
 namespace Microsoft.AspNetCore.ResponseCaching.Internal
 {
-    internal class ResponseCacheStream : Stream
+    internal class StreamWrapper : Stream
```
Tratcher: revert

Tratcher:
:shipit:

PR #30, opened 2016-09-08
Split headers body

JunTaoLuo:
- [x] Add an implementation of FastGuid
- [x] Use Guid based key generation
- [x] Update logic for normalizing and comparing cachedVaryBy entries 
- [x] Add an option to configure the size at which head and body are stored separately

cc @Tratcher @davidfowl

PR #31, opened 2016-09-09
Enumerable lookup keys

JunTaoLuo:
Adding support for multiple lookup keys. This will allow for scenarios such as varyby content encoding and using a cached response of a GET request to satisfy a HEAD response.

Pending #30

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/KeyProvider.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/KeyProvider.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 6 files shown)


Tratcher:
It may be less confusing if you did this as a PR targeting your other branch rather than dev.

src/Microsoft.AspNetCore.ResponseCaching/KeyProvider.cs
```diff
@@ -36,9 +36,19 @@ public KeyProvider(ObjectPoolProvider poolProvider, IOptions<ResponseCachingOpti
             _options = options.Value;
         }
 
+        public virtual IEnumerable<string> CreateLookupBaseKey(HttpContext httpContext)
+        {
+            yield return CreateStorageBaseKey(httpContext);
```
Tratcher: you may want to check this in the profiler. yield return is easy, but I don't know that it's efficient. Returning a string[1] would probably be more efficient.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -59,86 +59,103 @@ internal ResponseCachingState State
...
+                var cacheEntry = _cache.Get(baseKey);
+
+                if (cacheEntry is CachedVaryRules)
+                {
+                    // Request contains vary rules, recompute key(s) and try again
+                    State.CachedVaryRules = cacheEntry as CachedVaryRules;
 
-                        // Copy the cached response body
-                        if (body.Length > 0)
+                    foreach (var varyKey in _keyProvider.CreateLookupVaryKey(_httpContext, ((CachedVaryRules)cacheEntry).VaryRules))
```
BrennanConroy: Use `State.CachedVaryRules` instead of casting again?

Tratcher:
:shipit:

Issue #7, opened 2016-06-19
Cache chunked rather than whole byte[]


benaadams:
Cached data should be stored in < 85k chunks rather than as whole byte[] so they live in GC generations rather than LoH and can be cleaned without requiring a full Gen2

benaadams:
Fragmented stream example https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream

JunTaoLuo:
Sharding has been implemented in https://github.com/aspnet/ResponseCaching/pull/57 for `MemoryResponseCacheStore`.

benaadams:
@JunTaoLuo very nice, thank you!

PR #33, opened 2016-09-14
API updates

JunTaoLuo:
This contains many of the renames and reorganization that came up during the API review. 

cc @Tratcher @davidfowl


src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -60,58 +65,340 @@ public class ResponseCachingMiddleware
...
+                    if (body.Length > 0)
                     {
-                        // Subscribe to OnStarting event
-                        context.Response.OnStarting(OnStartingCallback, cachingContext);
+                        // Add a content-length if required
+                        if (response.ContentLength == null && StringValues.IsNullOrEmpty(response.Headers[HeaderNames.TransferEncoding]))
+                        {
+                            response.ContentLength = body.Length;
+                        }
+                        await response.Body.WriteAsync(body, 0, body.Length);
```
pranavkm: You could remove the `async` keyword on this method and just return the result here instead.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -60,58 +65,340 @@ public class ResponseCachingMiddleware
...
+        internal static async Task<bool> TryServeFromCacheAsync(ResponseCachingContext context)
+        {
+            foreach (var baseKey in context.KeyProvider.CreateLookupBaseKeys(context))
+            {
+                var cacheEntry = context.Cache.Get(baseKey);
+
+                if (cacheEntry is CachedVaryRules)
+                {
+                    // Request contains vary rules, recompute key(s) and try again
+                    context.CachedVaryRules = cacheEntry as CachedVaryRules;
```
pranavkm: C-style cast since this is already `CachedVaryRules`. Similarly other places in this method

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -60,58 +65,340 @@ public class ResponseCachingMiddleware
...
+            {
+                context.ShouldCacheResponse = true;
+                context.StorageBaseKey = context.KeyProvider.CreateStorageBaseKey(context);
+
+                // Create the cache entry now
+                var response = context.HttpContext.Response;
+                var varyHeaderValue = response.Headers[HeaderNames.Vary];
+                var varyParamsValue = context.HttpContext.GetResponseCachingFeature()?.VaryParams ?? StringValues.Empty;
+                context.CachedResponseValidFor = context.ResponseCacheControlHeaderValue.SharedMaxAge
+                    ?? context.ResponseCacheControlHeaderValue.MaxAge
```
pranavkm: Operator on the previous line: `context.ResponseCacheControlHeaderValue.SharedMaxAge ?? ...`

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -60,58 +65,340 @@ public class ResponseCachingMiddleware
...
+                    ?? (context.TypedResponseHeaders.Expires - context.ResponseTime)
+                    ?? TimeSpan.FromSeconds(10);
```
pranavkm: Should this be a const field on this type?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -60,58 +65,340 @@ public class ResponseCachingMiddleware
...
+                 context.TypedResponseHeaders.ContentLength == context.ResponseCacheStream.BufferedStream.Length))
+            {
+                if (context.ResponseCacheStream.BufferedStream.Length >= context.Options.MinimumSplitBodySize)
+                {
+                    // Store response and response body separately
+                    context.Cache.Set(context.StorageVaryKey ?? context.StorageBaseKey, context.CachedResponse, context.CachedResponseValidFor);
+
+                    var cachedResponseBody = new CachedResponseBody()
+                    {
+                        Body = context.ResponseCacheStream.BufferedStream.ToArray()
```
pranavkm: This might result in super large byte arrays going to the the large object heap. We use a "paged" buffer type in Mvc to avoid creating large strings, so a similar strategy might be nice here:

* Current code's here: https://github.com/aspnet/Mvc/blob/df845236bc9f4d4266fc2b124fcd3ddaf7507dcf/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
* Some refactoring that's in progress: https://github.com/aspnet/Mvc/blob/df845236bc9f4d4266fc2b124fcd3ddaf7507dcf/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedCharBuffer.cs

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -60,58 +65,340 @@ public class ResponseCachingMiddleware
...
+        // Normalize order and casing
+        internal static StringValues GetNormalizedStringValues(StringValues stringVales)
+        {
+            if (stringVales.Count == 1)
+            {
+                return new StringValues(stringVales.ToString().ToUpperInvariant());
+            }
+            else
+            {
+                var originalArray = stringVales.ToArray();
```
pranavkm: Isn't `StringValues` an `IList`?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
JunTaoLuo: Yea that's something that we plan to add to avoid the LOH. Tracked in https://github.com/aspnet/ResponseCaching/issues/7

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
JunTaoLuo: The method returns a bool to indicate whether a response is served. Not sure if I want to return the task with a continuation to return a bool here.

pranavkm: Gotcha. Missed the return statement.

Tratcher: no, it's `Task<bool>`

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
JunTaoLuo: Are you suggesting LINQ here?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -60,58 +65,340 @@ public class ResponseCachingMiddleware
...
+                // Check if any vary rules exist
+                if (!StringValues.IsNullOrEmpty(varyHeaderValue) || !StringValues.IsNullOrEmpty(varyParamsValue))
+                {
+                    // Normalize order and casing of vary by rules
+                    var normalizedVaryHeaderValue = GetNormalizedStringValues(varyHeaderValue);
+                    var normalizedVaryParamsValue = GetNormalizedStringValues(varyParamsValue);
+
+                    // Update vary rules if they are different
+                    if (context.CachedVaryRules == null ||
+                        !StringValues.Equals(context.CachedVaryRules.Params, normalizedVaryParamsValue) ||
```
pranavkm: Could we add an overload to `StringValues` that uses a `StringComparer`? Might remove the need to normalize these.

JunTaoLuo: We need to normalize order though, `StringValues` contains an underlying array. Do you recommend sorting the underlying array when a certain `StringComparer` is set? That modifies the original value so I'm not sure if that's a good idea.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
JunTaoLuo: Sort of. Not all `IList` functionalities are implemented. It's more like a readonly list. Why? Do you recommend using the IList.Sort() method? That's probably not really beneficial since the underlying data structure is an array so we'll just operate on that instead.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs
```diff
@@ -3,356 +3,119 @@
...
-
-        internal bool ResponseStarted { get; set; }
-
-        private Stream OriginalResponseStream { get; set; }
-
-        private ResponseCacheStream ResponseCacheStream { get; set; }
+        public HttpContext HttpContext { get; }
 
-        private IHttpSendFileFeature OriginalSendFileFeature { get; set; }
+        internal IResponseCache Cache { get; }
```
Tratcher: Why are these things all internal? There's nothing secret about them.

JunTaoLuo: Yea, but is there any reason to expose them here? Just resolve it from DI if you need it.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
pranavkm: Nevermind, I assumed this was the Linq ToArray. That said, `ToArray` returns the original array in the event you have more than one value, so you're overwriting values from the originalArray.

See: https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.Primitives/StringValues.cs#L113

JunTaoLuo: I'm creating a new array from the old one, see the next line?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
Tratcher: Option

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -208,7 +204,7 @@ public class ResponseCachingMiddleware
             return false;
         }
 
-        internal static bool ConditionalRequestSatisfied(ResponseCachingContext context)
+        internal bool ConditionalRequestSatisfied(ResponseCachingContext context)
```
Tratcher: This one could actually stay static, it doesn't use any member state.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -367,7 +363,7 @@ internal static void ShimResponseStream(ResponseCachingContext context)
             context.HttpContext.AddResponseCachingFeature();
         }
 
-        internal static void UnshimResponseStream(ResponseCachingContext context)
+        internal void UnshimResponseStream(ResponseCachingContext context)
```
Tratcher: This one too

Tratcher:
:shipit:

PR #34, opened 2016-09-14
Handle commas in header string values

JunTaoLuo:
We need to split each entry in the StringValues we get by commas for headers.

cc @BrennanConroy @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheMiddleware.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheMiddleware.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
@@ -211,7 +212,7 @@ namespace Microsoft.AspNetCore.ResponseCaching
 
                 // Create the cache entry now
                 var response = context.HttpContext.Response;
-                var varyHeaderValue = response.Headers[HeaderNames.Vary];
+                var varyHeaderValue = new StringValues(response.Headers.GetCommaSeparatedValues(HeaderNames.Vary));
                 var varyParamsValue = context.HttpContext.GetResponseCacheFeature()?.VaryByParams ?? StringValues.Empty;
                 context.CachedResponseValidFor = context.ResponseCacheControlHeaderValue.SharedMaxAge ??
                     context.ResponseCacheControlHeaderValue.MaxAge ??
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -375,8 +376,44 @@ internal static bool ConditionalRequestSatisfied(ResponseCachingContext context)
...
+            if (!containsComma)
+            {
+                return GetOrderCasingNormalizedStringValues(stringValues);
+            }
+            else
+            {
+                var headers = new List<string>(stringValues.Count);
+                foreach (var value in stringValues)
+                {
+                    foreach (var header in value.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries))
```
JunTaoLuo: Not sure if this is the most performant way to do this

test/Microsoft.AspNetCore.ResponseCaching.Tests/ResponseCachingMiddlewareTests.cs
```diff
@@ -554,25 +554,68 @@ public void FinalizeCachingHeaders_StoresCachedResponse_InState()
...
         }
+
+        [Fact]
+        public void GetOrderCasingNormalizedStringValues_PreservesCommas()
+        {
+            var originalStrings = new StringValues(new[] { "STRINGA, STRINGB" });
+
+            var normalizedStrings = ResponseCachingMiddleware.GetOrderCasingNormalizedStringValues(originalStrings);
+
+            Assert.Equal(originalStrings, normalizedStrings);
```
BrennanConroy: Is this a string.Equals and the other tests are StringValues.Equals?

JunTaoLuo: This is still a StringValues.Equals, even though both StringValues only contain one string and calls string.Equals internally. What's the concern here?

BrennanConroy: The example here https://github.com/aspnet/ResponseCaching/pull/34/files#diff-80295bfdc844391de2ffaeba04ef7843R613 shows a similar scenario split into 2 values in the StringValues, so this one doesn't get split into 2?

JunTaoLuo: Correct, we don't want this version to split.

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
Tratcher: IHeaderDictionary.GetCommaSepratedValues

JunTaoLuo:
update @davidfowl

Tratcher:
:shipit:

PR #35, opened 2016-09-14
API review renames and updates continued

JunTaoLuo:
- Consolidate base key to be singular

cc @BrennanConroy @glennc @Tratcher


src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.AspNetCore.ResponseCaching.Internal
 {
-    internal class DistributedResponseCache : IResponseCache
+    internal class DistributedResponseCache : IResponseCacheStore
```
Tratcher: DistributedResponseCacheStore.

Why is this internal? recommend public in the Internal namespace.

src/Microsoft.AspNetCore.ResponseCaching/Internal/MemoryResponseCacheStore.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.AspNetCore.ResponseCaching.Internal
 {
-    internal class MemoryResponseCache : IResponseCache
+    internal class MemoryResponseCache : IResponseCacheStore
```
Tratcher: Same

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCacheStore.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.AspNetCore.ResponseCaching
 {
-    public interface IResponseCache
+    public interface IResponseCacheStore
     {
         object Get(string key);
```
Tratcher: Side note: Get and Set will likely need to be async for the distributed/redis scenarios

JunTaoLuo: yea, will probably investigate this as part of sharding.

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
JunTaoLuo: Wow... I changed the filename but forgot the type.

src/Microsoft.AspNetCore.ResponseCaching/Internal/MemoryResponseCacheStore.cs
```diff
@@ -6,11 +6,11 @@
 
 namespace Microsoft.AspNetCore.ResponseCaching.Internal
 {
-    internal class MemoryResponseCache : IResponseCache
+    public class MemoryResponseCacheStore : IResponseCacheStore
```
davidfowl: InMemory?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheFeature.cs
```diff
@@ -6,8 +6,8 @@
 namespace Microsoft.AspNetCore.ResponseCaching
 {
     // TODO: Temporary interface for endpoints to specify options for response caching
-    public class ResponseCachingFeature
+    public class ResponseCacheFeature
```
davidfowl: Why is this public? Who uses it?

Tratcher: This is the feature for apps to set custom vary-by's. It has to be a feature because there's no header for this one.

src/Microsoft.AspNetCore.ResponseCaching/Internal/MemoryResponseCacheStore.cs
JunTaoLuo: I thought about that too but it seems we call it `MemoryCache` everywhere. The type is called `MemoryCache`, the interface is `IMemoryCache` and most members are called `MemoryCache` https://github.com/aspnet/Mvc/blob/d4c7dc3a834edc507865816a59960c5b31f9b10c/src/Microsoft.AspNetCore.Mvc.TagHelpers/DistributedCacheTagHelper.cs#L47,
https://github.com/aspnet/EntityFramework/blob/e60330ef54e0ab02eb4010f91b252e2ed67177bd/src/Microsoft.EntityFrameworkCore/Internal/IDbContextServices.cs#L51,

Also I have not seen anything called `InMemoryCache` and I don't think that a pattern we want to start.

PR #47, opened 2016-09-16
Make IResponseCacheStore APIs async

JunTaoLuo:


JunTaoLuo:
I think this is ready cc @BrennanConroy @Tratcher @davidfowl

src/Microsoft.AspNetCore.ResponseCaching/CacheEntry/CacheEntryContext.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.ResponseCaching
+{
+    public class KeyValuePair
```
JunTaoLuo: Might need better name

BrennanConroy: Curious, why not use https://msdn.microsoft.com/en-us/library/5tbh8a42(v=vs.110).aspx

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCacheStore.cs
```diff
@@ -2,13 +2,14 @@
...
+using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.ResponseCaching
 {
     public interface IResponseCacheStore
     {
-        object Get(string key);
-        void Set(string key, object entry, TimeSpan validFor);
-        void Remove(string key);
+        Task PopulateValueAsync(KeyValuePair keyValuePair);
```
JunTaoLuo: Would love to do that except you can't do that on an async method.

src/Microsoft.AspNetCore.ResponseCaching/CacheEntry/CacheEntryContext.cs
JunTaoLuo: Need a reference type here so I can populate the value. Won't work with a struct.

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCacheStore.cs
Tratcher: As opposed to returning `Task<object>`?

Tratcher: Didn't you want to stop using object here anyways?

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCacheStore.cs
```diff
@@ -2,13 +2,14 @@
...
+        Task SetAsync(KeyValuePair keyValuePair, TimeSpan validFor);
+        Task RemoveAsync(KeyValuePair keyValuePair);
```
Tratcher: just key

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCacheStore.cs
```diff
@@ -2,13 +2,14 @@
...
 
 namespace Microsoft.AspNetCore.ResponseCaching
 {
     public interface IResponseCacheStore
     {
-        object Get(string key);
-        void Set(string key, object entry, TimeSpan validFor);
-        void Remove(string key);
+        Task PopulateValueAsync(KeyValuePair keyValuePair);
+        Task SetAsync(KeyValuePair keyValuePair, TimeSpan validFor);
```
Tratcher: separate key and value parameters

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCacheStore.cs
Tratcher: Needs numbers to show that KVP is significantly lighter than `Task<T>`

davidfowl: Task<T> is uncachable. That's the problem. It'll make the sync code path much worse.

JunTaoLuo: I'll compare the KVP vs the Task<object> implementation from the first commit and see what, if any, differences there are.

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCacheStore.cs
JunTaoLuo: Originally wanted to keep some consistency between the IResponseCacheStore methods but will revisit once we decide which implementation to go with.

JunTaoLuo:
image
So here's what a quick profiling of processing time yielded. The difference between the two implementations is very small. The KVP implementation seems slightly faster for IMemoryCache and the Task<object> implementation is slightly faster for IDistributedCache. I'll also look at memory usage and post what I find.

JunTaoLuo:
image
Similar patterns for memory traffic, KVP is better for IMemoryCache and Task<object> is better for IDistributedCache

Tratcher:
So no practical difference.

davidfowl:
What did the benchmark code look like?

JunTaoLuo:
The consensus is to use the Task<object> implementation, :up::date:

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -20,40 +24,46 @@ public DistributedResponseCacheStore(IDistributedCache cache)
...
+                return _cache.SetAsync(
                     key,
                     CacheEntrySerializer.Serialize(entry),
                     new DistributedCacheEntryOptions()
                     {
                         AbsoluteExpirationRelativeToNow = validFor
                     });
             }
-            catch { }
+            catch
```
Tratcher: await or you'll miss half of the exceptions

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -20,40 +24,46 @@ public DistributedResponseCacheStore(IDistributedCache cache)
...
             }
         }
 
-        public void Remove(string key)
+        public Task RemoveAsync(string key)
         {
             try
             {
-                _cache.Remove(key);
+                return _cache.RemoveAsync(key);
```
Tratcher: await

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCacheStore.cs
```diff
@@ -2,13 +2,14 @@
...
 namespace Microsoft.AspNetCore.ResponseCaching
 {
     public interface IResponseCacheStore
     {
-        object Get(string key);
-        void Set(string key, object entry, TimeSpan validFor);
-        void Remove(string key);
+        Task<object> GetAsync(string key);
+        Task SetAsync(string key, object entry, TimeSpan validFor);
+        Task RemoveAsync(string key);
```
Tratcher: RemoveAsync isn't used anywhere yet, correct?

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
davidfowl: hmmm @JunTaoLuo did you profile without awaiting?

src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCacheStore.cs
JunTaoLuo: The original idea was to make use of this for invalidation. I'm not sure if that's still on the table but this would be needed if we eventually want to add it.

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
JunTaoLuo: I didn't await here. Was awaiting in the middleware.

JunTaoLuo: image

Updated the measurements after the change. The processing time is now favouring KVP in all cases but the difference is so small, it's within margin of error.

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
JunTaoLuo: updated

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -20,32 +24,32 @@ public DistributedResponseCacheStore(IDistributedCache cache)
...
         {
             try
             {
-                return CacheEntrySerializer.Deserialize(_cache.Get(key));
+                return CacheEntrySerializer.Deserialize(await _cache.GetAsync(key));
             }
             catch
             {
-                return null;
+                return NullCompletedTask;
```
Tratcher: double wrapped

PR #48, opened 2016-09-17
Short circuit to avoid DateTimeOffset parsing

BrennanConroy:
@JunTaoLuo

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheMiddleware.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheMiddleware.cs
@@ -355,7 +355,7 @@ namespace Microsoft.AspNetCore.ResponseCaching
                     }
                 }
             }
-            else if ((cachedResponseHeaders.LastModified ?? cachedResponseHeaders.Date) <= context.TypedRequestHeaders.IfUnmodifiedSince)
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
:shipit:

src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheMiddleware.cs
```diff
@@ -355,7 +355,7 @@ internal static bool ConditionalRequestSatisfied(ResponseCacheContext context)
                     }
                 }
             }
-            else if ((cachedResponseHeaders.LastModified ?? cachedResponseHeaders.Date) <= context.TypedRequestHeaders.IfUnmodifiedSince)
+            else if (context.TypedRequestHeaders.IfUnmodifiedSince != null && (cachedResponseHeaders.LastModified ?? cachedResponseHeaders.Date) <= context.TypedRequestHeaders.IfUnmodifiedSince)
```
Tratcher: Remember that ResponseHeaders properties have no caching, they parse every time you access them

Tratcher: and RequestHeaders...

JunTaoLuo: More cached fields on the context then.

PR #49, opened 2016-09-19
Make freshness more strict

JunTaoLuo:
max-age=0 should never be satisfied given that utcnow only has a 10-15ms resolution
also avoid duplicate work when parsing typed headers.

cc @BrennanConroy @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheContext.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheContext.cs
@@ -19,6 +19,10 @@ namespace Microsoft.AspNetCore.ResponseCaching
         private ResponseHeaders _responseHeaders;
         private CacheControlHeaderValue _requestCacheControl;
         private CacheControlHeaderValue _responseCacheControl;
+        private DateTimeOffset? _responseDate;
+        private bool _parsedResponseDate;
+        private DateTimeOffset? _responseExpires;
+        private bool _parsedResponseExpires;
 
         internal ResponseCacheContext(
             HttpContext httpContext)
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheContext.cs
```diff
@@ -19,6 +19,10 @@ public class ResponseCacheContext
         private ResponseHeaders _responseHeaders;
         private CacheControlHeaderValue _requestCacheControl;
         private CacheControlHeaderValue _responseCacheControl;
+        private DateTimeOffset? _responseDate;
+        private bool _parsedResponseDate;
```
BrennanConroy: These bools aren't necessary, just check for null

JunTaoLuo: what if the responsedate is `null` when parsed?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheMiddleware.cs
```diff
@@ -238,16 +238,18 @@ internal void FinalizeCacheHeaders(ResponseCacheContext context)
                 }
 
                 // Ensure date header is set
-                if (context.TypedResponseHeaders.Date == null)
+                if (context.ResponseDate == null)
                 {
-                    context.TypedResponseHeaders.Date = context.ResponseTime;
+                    context.ResponseDate = context.ResponseTime;
+                    // Setting the date on the raw response headers.
+                    context.TypedResponseHeaders.Date = context.ResponseDate;
```
BrennanConroy: Half the point was to remove this line since it is unnecessary and a lot of extra work

src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheMiddleware.cs
```diff
@@ -238,16 +238,18 @@ internal void FinalizeCacheHeaders(ResponseCacheContext context)
                 }
 
                 // Ensure date header is set
-                if (context.TypedResponseHeaders.Date == null)
+                if (context.ResponseDate == null)
```
Tratcher: `!...HasValue`

Tratcher: Yeah, use HasValue for nullables rather than comparing to null.

BrennanConroy:
Looks good except for my comment (for perf!)

JunTaoLuo:
offline: no perf issue so going to merge now.

PR #50, opened 2016-09-20
API updates

JunTaoLuo:
- Internalize properties on ResponseCacheContext unless required by extension points
 - Rename VaryByParams -> VaryByQueryKeys

cc @Tratcher @glennc

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheContext.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheContext.cs
@@ -32,9 +32,9 @@ namespace Microsoft.AspNetCore.ResponseCaching
 
         public HttpContext HttpContext { get; }
 
-        public DateTimeOffset ResponseTime { get; internal set; }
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheContext.cs
```diff
@@ -32,21 +32,21 @@ public class ResponseCacheContext
 
         public HttpContext HttpContext { get; }
 
-        public bool ShouldCacheResponse { get; internal set; }
+        public DateTimeOffset ResponseTime { get; internal set; }
```
JunTaoLuo: Should we make this value type nullable? Otherwise the default value might be confusing. Same with CachedEntryAge

Tratcher: Makes sense

JunTaoLuo:
merged to dev in https://github.com/aspnet/ResponseCaching/commit/5e12a103a4c204d4b3fb68d07e43ce2299b29cf9

PR #51, opened 2016-09-21
Cache HeaderDictionary in Serializer

BrennanConroy:
Saves ~280 bytes per request which is ~3.8% of allocations.

No noticeable perf changes in single page plaintext testing.

@JunTaoLuo @Tratcher @davidfowl @rynowak


Tratcher:
no no no no no no no....
This doesn't properly manage the lifetime of the cached object.

Issue #42, opened 2016-09-15
Logging


JunTaoLuo:
Add logging to indicate what the response cache is doing.

PR #56, opened 2016-09-22
Sharding (WIP)

JunTaoLuo:


src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -4,28 +4,46 @@
...
             try
             {
-                return CacheEntrySerializer.Deserialize(await _cache.GetAsync(key));
+                var entry = ResponseCacheEntrySerializer.Deserialize(await _cache.GetAsync(key));
+
+                var cachedResponse = entry as CachedResponse;
+                if (cachedResponse != null)
+                {
+                    var shardCount = (cachedResponse.Body.Length + _options.CachedBodyShardSize - 1) / _options.CachedBodyShardSize;
+                    var copyOnlyDistributedCacheStream = (CopyOnlyDistributedCacheStream)cachedResponse.Body;
```
Tratcher: This is strange. Why even initialize the Body in the deserializer if you can't fully configure it?

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -4,28 +4,46 @@
...
         {
             try
             {
-                return CacheEntrySerializer.Deserialize(await _cache.GetAsync(key));
+                var entry = ResponseCacheEntrySerializer.Deserialize(await _cache.GetAsync(key));
+
+                var cachedResponse = entry as CachedResponse;
+                if (cachedResponse != null)
+                {
+                    var shardCount = (cachedResponse.Body.Length + _options.CachedBodyShardSize - 1) / _options.CachedBodyShardSize;
```
Tratcher: Why not just store the shard count? there's no reason for you to be constrained by the CachedResponse object signature internally, make your own local structure to serialize.

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -42,17 +60,51 @@ public DistributedResponseCacheStore(IDistributedCache cache)
...
 
-        public async Task SetAsync(string key, object entry, TimeSpan validFor)
+        public async Task SetAsync(string key, IResponseCacheEntry entry, TimeSpan validFor)
         {
             try
             {
+                var cachedResponse = entry as CachedResponse;
+                if (cachedResponse != null)
+                {
+                    cachedResponse.BodyKeyPrefix = FastGuid.NewGuid().IdString;
```
Tratcher: BodyKeyPrefix is an implementation detail of the store, it shouldn't be on CachedResponse

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -42,17 +60,51 @@ public DistributedResponseCacheStore(IDistributedCache cache)
...
                     });
+
+                if (cachedResponse != null)
+                {
+                    var shardCount = (cachedResponse.Body.Length + _options.CachedBodyShardSize - 1) / _options.CachedBodyShardSize;
+
+                    for (int i = 0; i < shardCount; i++)
+                    {
+                        // TODO: doesn't need a new shard every time?
+                        var shard = new byte[_options.CachedBodyShardSize];
```
Tratcher: buffer pool

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
muratg: @Tratcher Doesn't the number of shards depend on the body size? So it would change response to response.

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
muratg: (nit: space after cast)

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -42,17 +60,51 @@ public DistributedResponseCacheStore(IDistributedCache cache)
             catch { }
```
muratg: Is there any exception that we'd actually like catch for logging purposes? Or do we not care at all since this is `Remove`?

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -42,17 +60,51 @@ public DistributedResponseCacheStore(IDistributedCache cache)
...
+                    var shardCount = (cachedResponse.Body.Length + _options.CachedBodyShardSize - 1) / _options.CachedBodyShardSize;
+
+                    for (int i = 0; i < shardCount; i++)
+                    {
+                        // TODO: doesn't need a new shard every time?
+                        var shard = new byte[_options.CachedBodyShardSize];
+                        var bytesRead = cachedResponse.Body.Read(shard, 0, _options.CachedBodyShardSize);
+
+                        // The last shard may not be full
+                        if (bytesRead != _options.CachedBodyShardSize)
```
muratg: Can this happen due to the underlying store for any of the other shards (not just the last one)? Perhaps, for example, reads not completing fully. 

Also, could `bytesRead` be ever 0?

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
```diff
@@ -42,17 +60,51 @@ public DistributedResponseCacheStore(IDistributedCache cache)
...
                     new DistributedCacheEntryOptions()
                     {
                         AbsoluteExpirationRelativeToNow = validFor
                     });
+
+                if (cachedResponse != null)
+                {
+                    var shardCount = (cachedResponse.Body.Length + _options.CachedBodyShardSize - 1) / _options.CachedBodyShardSize;
+
+                    for (int i = 0; i < shardCount; i++)
```
muratg: nit: `var i`.

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheEntrySerializer.cs
```diff
@@ -2,16 +2,18 @@
...
-    internal static class CacheEntrySerializer
+    internal static class ResponseCacheEntrySerializer
     {
+        private static readonly List<byte[]> EmptyList = new List<byte[]>();
         private const int FormatVersion = 1;
 
-        public static object Deserialize(byte[] serializedEntry)
+        public static IResponseCacheEntry Deserialize(byte[] serializedEntry)
         {
             if (serializedEntry == null)
```
muratg: Could this be empty, i.e. a 0-length array?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheOptions.cs
```diff
@@ -14,14 +14,15 @@ public class ResponseCacheOptions
         public long MaximumCachedBodySize { get; set; } = 1024 * 1024;
 
         /// <summary>
-        /// <c>true</c> if request paths are case-sensitive; otherwise <c>false</c>. The default is to treat paths as case-insensitive.
+        /// The shard size for the response body in bytes. The default is set to 1 KB.
         /// </summary>
-        public bool UseCaseSensitivePaths { get; set; } = false;
+        // TODO: Setting to 1KB for testing, 75 KB is more reasonable.
+        public int CachedBodyShardSize { get; set; } = 10;
```
muratg: Unless I'm missing something, this is not 1KB 😄

src/Microsoft.AspNetCore.ResponseCaching/Streams/CopyOnlyDistributedCacheStream.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public override bool CanSeek => false;
+
+        public override bool CanWrite => false;
+
+        public override long Length => _length;
+
+        public override long Position
+        {
+            get { return 0; }
```
muratg: Just to make Stream happy?

src/Microsoft.AspNetCore.ResponseCaching/Streams/CopyOnlyDistributedCacheStream.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public override long Seek(long offset, SeekOrigin origin) => 0;
+
+        public override void SetLength(long value) { }
+
+        public override void Write(byte[] buffer, int offset, int count) { }
+
+        public override async Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
+        {
+            // TODO: Check validity of parameters
```
muratg: +1

src/Microsoft.AspNetCore.ResponseCaching/Streams/ResponseCacheStream.cs
```diff
@@ -12,37 +13,68 @@ internal class ResponseCacheStream : Stream
...
         {
             _innerStream = innerStream;
             _maxBufferSize = maxBufferSize;
+            _bufferShardSize = bufferShardSize;
+            _shards = new List<byte[]>();
+            _bufferStream = new MemoryStream();
         }
 
-        public MemoryStream BufferedStream { get; } = new MemoryStream();
+        public int BufferShardSize => _bufferShardSize;
```
muratg: `ShardBufferSize` perhaps?  (What do you think, @Tratcher?)

src/Microsoft.AspNetCore.ResponseCaching/Streams/ResponseCacheStream.cs
```diff
@@ -12,37 +13,68 @@ internal class ResponseCacheStream : Stream
...
 
-        public override long Length => _innerStream.Length;
+        public override long Length => _bufferedBytes;
 
         public override long Position
         {
             get { return _innerStream.Position; }
-            set { _innerStream.Position = value; }
+            set
+             {
```
muratg: nit: extra space

src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheOptions.cs
JunTaoLuo: Was 10k until I wanted to force sharding in testing. I often use GUIDs for response bodies so I needed a smaller shard size. That said, I only opened this PR to show what sharding with streams look like, there are tons of minor details that still need to be polished and it's not really ready for final reviewing just yet.

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
JunTaoLuo: We can skip the calculation on the GetAsync by storing the result of the calculation during SetAsync

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
JunTaoLuo: Curious if that's the style we have everywhere else? I have always skipped the space and I see it elsewhere: https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.Primitives/StringValues.cs#L390-L400

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
JunTaoLuo: We will probably log exceptions, such as when the connection is unreliable. Adding logging is a todo item https://github.com/aspnet/ResponseCaching/issues/42

src/Microsoft.AspNetCore.ResponseCaching/Internal/DistributedResponseCacheStore.cs
JunTaoLuo: Hmm this is unlikely since this is reading from the buffer stream and that's in memory. Though when we add buffering from disk, that might be a concern. We will never get a bytesRead of 0 given the current in-memory buffering. But again, error handling will likely need to be added when we implement buffering from disk.

src/Microsoft.AspNetCore.ResponseCaching/Streams/ResponseCacheStream.cs
JunTaoLuo: `ShardSize` is probably sufficient?

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheEntrySerializer.cs
JunTaoLuo: Of course, but that would be an invalid entry and the exception will be handled by the DistributedResponseCacheStore.

Tratcher:
Design question: When sharding into a distributed cache, how do you deal with the race condition between adding an entry to the cache and having it retrieved and served before all of the body is cached? This may result in a lot of failed truncated responses.  One option is to cache the body segments in reverse order but this creates other issues.

Related: Cache entries use reproducible keys and are updated by being overwritten, but body shards use GUIDs and are not overwritten. If an entry is frequently updated the cache will quickly fill up with obsolete body shards.

Tratcher:
Trying to manually shard into an IDistributedCache is conceptually flawed, there's no atomicity. We're going to have tearing and leaks, reducing site reliability. Any sharding needs to be implemented directly within a cache where it can correlate the fragments and return, update or remove them together.

JunTaoLuo:
I agree that there are limitations for sharding for IDistributedCache based stores I still think it's possible, though possibly less efficient. One way is to buffer all shards in memory and validate all shards are received before serving the response and this should be reasonable as long as the response isn't too large. We can also remove shards of the old response when overwriting to reduce obsolete body shards. Finally, races between adding entries and retrieving them can be mitigated by writing the response entry after all shards are written. Note that these solutions come with their own efficiency and reliability problems as well.

Regardless, I do agree that we are doing a lot of extra work to accommodate the limitations of the underlying abstraction and potentially reducing the efficiency and reliability of the cache. I'll probably disable sharding for distributed caches and we can revisit later on since this is an internal implementation detail.

PR #57, opened 2016-09-26
Sharding

JunTaoLuo:
cc @Tratcher 

Still working on polishing, a few TODOs and adding tests. But otherwise ready for review.


src/Microsoft.AspNetCore.ResponseCaching/ResponseOutput/ReadOnlyShardStream.cs
```diff
@@ -0,0 +1,218 @@
...
+    internal class ReadOnlyShardStream : Stream
+    {
+        private readonly List<byte[]> _shards;
+        private readonly long _length;
+        private readonly int _shardSize;
+        private int _shardIndex;
+        private int _shardOffset;
+        private long _position;
+
+        internal ReadOnlyShardStream(List<byte[]> shards, int shardSize, long length)
```
JunTaoLuo: Assuming shardSize and length are not lies. Given that this is internal though, I have not added logic to check these values.

src/Microsoft.AspNetCore.ResponseCaching/Internal/MemoryCachedResponse.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public DateTimeOffset Created { get; set; }
+
+        public int StatusCode { get; set; }
+
+        public IHeaderDictionary Headers { get; set; } = new HeaderDictionary();
+
+        public List<byte[]> Shards { get; set; }
+
+        public int ShardSize { get; set; }
```
Tratcher: Why do you need the shard size if you have the  shards?

src/Microsoft.AspNetCore.ResponseCaching/Internal/MemoryResponseCacheStore.cs
```diff
@@ -3,46 +3,86 @@
...
             {
                 throw new ArgumentNullException(nameof(cache));
             }
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
 
             _cache = cache;
+            _options = options.Value;
```
JunTaoLuo: Same option in `ResponseCacheOptions` configures the shard size of the buffer in `ResponseCacheStream` and `ReadOnlyShardStream`.

src/Microsoft.AspNetCore.ResponseCaching/Internal/MemoryCachedResponse.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    internal class MemoryCachedResponse
+    {
+        public DateTimeOffset Created { get; set; }
+
+        public int StatusCode { get; set; }
+
+        public IHeaderDictionary Headers { get; set; } = new HeaderDictionary();
+
+        public List<byte[]> Shards { get; set; }
```
Tratcher: BodySegments. Maybe `List<ArraySegment<byte>>` for buffer pooling purposes.

src/Microsoft.AspNetCore.ResponseCaching/Internal/MemoryCachedResponse.cs
JunTaoLuo: you are looking at outdate diff haha

src/Microsoft.AspNetCore.ResponseCaching/Internal/MemoryCachedResponse.cs
JunTaoLuo: Will need to come up with a consistent terminology since the naming is exposed via options. I might leave the pooling optimization to another PR. This one is already getting laggy on github.

src/Microsoft.AspNetCore.ResponseCaching/ResponseOutput/ReadOnlyShardStream.cs
```diff
@@ -0,0 +1,216 @@
...
+            }
+
+            if (_shardOffset == _shards[_shardIndex].Length)
+            {
+                // Move to the next shard
+                _shardIndex++;
+                _shardOffset = 0;
+            }
+
+            var byteRead = _shards[_shardIndex][_shardOffset];
```
BrennanConroy: _shards.Count check?

src/Microsoft.AspNetCore.ResponseCaching/ResponseOutput/WriteOnlyShardStream.cs
```diff
@@ -0,0 +1,168 @@
...
+        public override void SetLength(long value)
+        {
+            throw new NotSupportedException("The stream does not support seeking.");
+        }
+
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            if (!CanWrite)
+            {
+                throw new InvalidOperationException("The stream has been closed for writing.");
```
Tratcher: ObjectDisposedException?

src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheOptions.cs
```diff
@@ -9,19 +9,20 @@ namespace Microsoft.AspNetCore.Builder
     public class ResponseCacheOptions
     {
         /// <summary>
-        /// The largest cacheable size for the response body in bytes. The default is set to 1 MB.
+        /// The shard size for buffering the response body in bytes. The default is set to 80 KB.
         /// </summary>
-        public long MaximumCachedBodySize { get; set; } = 1024 * 1024;
+        // TODO: Setting to 10 for testing, 80 KB is more reasonable.
+        public int BodyShardSize { get; set; } = 10;
```
JunTaoLuo: Well it avoids the LOH in terms of allocations.

src/Microsoft.AspNetCore.ResponseCaching/ResponseOutput/ResponseCacheStream.cs
```diff
@@ -12,37 +12,56 @@ internal class ResponseCacheStream : Stream
...
+        }
+
+        internal Stream GetShardStream()
+        {
+            if (_readOnlyStream == null)
+            {
+                _readOnlyStream = new ReadOnlyShardStream(_writeOnlyStream.Shards, _writeOnlyStream.Length);
+
+                // TODO: clean up the write stream
+                _writeOnlyStream = null;
```
JunTaoLuo: Yea, that was just a placeholder for actually cleaning it up properly

src/Microsoft.AspNetCore.ResponseCaching/ResponseOutput/ReadOnlyShardStream.cs
```diff
@@ -0,0 +1,216 @@
...
+#endif
+        {
+            if (asyncResult == null)
+            {
+                throw new ArgumentNullException(nameof(asyncResult));
+            }
+            return ((Task<int>)asyncResult).GetAwaiter().GetResult();
+        }
+
+        public override long Seek(long offset, SeekOrigin origin)
```
JunTaoLuo: Let's not implement something until we need it

src/Microsoft.AspNetCore.ResponseCaching/ResponseOutput/StreamUtilities.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
+{
+    public class StreamUtilities
```
JunTaoLuo: yea... that's supposed to be internal

src/Microsoft.AspNetCore.ResponseCaching/ResponseOutput/ReadOnlyShardStream.cs
```diff
@@ -0,0 +1,216 @@
...
+        {
+            get
+            {
+                return _position;
+            }
+            set
+            {
+                if (value != 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(value), value, $"{nameof(Position)} can only be set to 0.");
```
JunTaoLuo: Not going to implement random access until we need it

src/Microsoft.AspNetCore.ResponseCaching/ResponseOutput/ResponseCacheStream.cs
```diff
@@ -12,37 +12,56 @@ internal class ResponseCacheStream : Stream
...
+            set
+            {
+                DisableBuffering();
+                _innerStream.Position = value;
+            }
+        }
+
+        internal Stream GetShardStream()
+        {
+            if (_readOnlyStream == null)
```
JunTaoLuo: I don't expect it to be called multiple times. But if we did, this is actually wrong since we want to recreate a new ReadOnlyShardStream since reads are stateful.

src/Microsoft.AspNetCore.ResponseCaching/ResponseOutput/ResponseCacheStream.cs
```diff
@@ -12,37 +12,56 @@ internal class ResponseCacheStream : Stream
...
+        internal bool BufferingEnabled { get; private set; } = true;
 
         public override bool CanRead => _innerStream.CanRead;
 
         public override bool CanSeek => _innerStream.CanSeek;
 
         public override bool CanWrite => _innerStream.CanWrite;
 
-        public override long Length => _innerStream.Length;
+        public override long Length => _writeOnlyStream.Length;
```
JunTaoLuo: I think this should represent the length of the stream we have buffered. We don't really need to know the length of the underlying stream.

Tratcher: Except ResponseCacheStream is exposed directly to the application. If the underlying stream is tracking length then you'll return the incorrect answer in scenarios where buffering is disabled.

JunTaoLuo: Ah, that makes sense.

src/Microsoft.AspNetCore.ResponseCaching/ResponseOutput/WriteOnlyShardStream.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
+{
+    internal class WriteOnlyShardStream : Stream
+    {
+        private readonly List<byte[]> _shards = new List<byte[]>();
+        private readonly MemoryStream _bufferStream = new MemoryStream();
+        private readonly int _shardSize;
```
JunTaoLuo: Going to set a static limit of 84k right now. Can make this configurable in the future if the need arises.

src/Microsoft.AspNetCore.ResponseCaching/Internal/MemoryCachedResponse.cs
JunTaoLuo: `*Shard` => `*Segment`?

Tratcher: BodySegment

src/Microsoft.AspNetCore.ResponseCaching/Internal/MemoryResponseCacheStore.cs
```diff
@@ -22,27 +21,60 @@ public MemoryResponseCacheStore(IMemoryCache cache)
...
-                entry,
-                new MemoryCacheEntryOptions()
-                {
-                    AbsoluteExpirationRelativeToNow = validFor
-                });
-            return TaskCache.CompletedTask;
+            if (entry is CachedResponse)
+            {
+                var cachedResponse = (CachedResponse)entry;
+                var shardStream = new WriteOnlyShardStream(StreamUtilities.BodyShardSize);
```
Tratcher: SegmentWriteStream?

src/Microsoft.AspNetCore.ResponseCaching/Streams/WriteOnlyShardStream.cs
```diff
@@ -0,0 +1,215 @@
...
+            }
+            if (offset < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(offset), offset, "Non-negative number required.");
+            }
+            if (count < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(count), count, "Non-negative number required.");
+            }
+            if (buffer.Length < offset + count)
```
Tratcher: Always use `count > length - offset` to avoid integer overflows in `offset + count`.

src/Microsoft.AspNetCore.ResponseCaching/Streams/ReadOnlySegmentStream.cs
```diff
@@ -0,0 +1,237 @@
...
+        {
+            if (buffer == null)
+            {
+                throw new ArgumentNullException(nameof(buffer));
+            }
+            if (offset < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(offset), offset, "Non-negative number required.");
+            }
+            if (count < 0 )
```
Tratcher: <=, Read of count == 0 is invalid (but it's valid for Write)

src/Microsoft.AspNetCore.ResponseCaching/Streams/ReadOnlySegmentStream.cs
```diff
@@ -0,0 +1,237 @@
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(count), count, "Non-negative number required.");
+            }
+            if (buffer.Length < offset + count)
```
Tratcher: same as write stream

src/Microsoft.AspNetCore.ResponseCaching/Streams/WriteOnlySegmentStream.cs
```diff
@@ -0,0 +1,215 @@
...
+                }
+
+                if (disposing)
+                {
+                    _segments.Clear();
+                    DisposeMemoryStream();
+                }
+
+                _disposed = true;
+                _closed = true;
```
Tratcher: What's the difference between closed and disposed?

JunTaoLuo: Closed means closed for writing, disposed is when the streams has been disposed segments are cleared.

src/Microsoft.AspNetCore.ResponseCaching/Streams/StreamUtilities.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
+{
+    internal static class StreamUtilities
+    {
+        /// <summary>
+        /// The segment size for buffering the response body in bytes. The default is set to 84 KB.
+        /// </summary>
+        // Internal for testing
+        internal static int BodySegmentSize { get; set; } = 84 * 1024;
```
khellang: Is there a reason this was set to 86016 (which is just above the LOH limit of 85K)?

The default buffer size for `CopyTo{Async}` [is set to 81920](https://github.com/dotnet/coreclr/blob/b089964f6546a6473e0f5d90d0cae49f22c57455/src/mscorlib/src/System/IO/Stream.cs#L39) which is the largest multiple of 4096 that is smaller than the large object heap threshold. Shouldn't this follow the same limit?

JunTaoLuo: Yes it should.

JunTaoLuo: https://github.com/aspnet/ResponseCaching/pull/146

PR #58, opened 2016-09-28
Remove iresponsecachekeyprovider

JunTaoLuo:
cc @Tratcher @BrennanConroy


src/Microsoft.AspNetCore.ResponseCaching/Interfaces/IResponseCacheKeyProvider.cs
```diff
@@ -1,31 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System.Collections.Generic;
-
-namespace Microsoft.AspNetCore.ResponseCaching
-{
-    public interface IResponseCacheKeyProvider
```
Tratcher: Do you need to remove this vs just making it internal?

JunTaoLuo: I don't think the interface is valuable to keep? We only have one internal implementation of it anyways.

Tratcher: If it's part of the future plans there's no reason to delete it, we're just making it internal because we're worried it's not done yet.

JunTaoLuo: There is discussion with whether it should be string based at all, and hashing of keys. I expect the interface to change dramatically anyway when we revisit.

Tratcher: Wouldn't that just change the return value from string to something else?

JunTaoLuo:
Discussed offline, though the interface will likely change significantly, we'll retain it to reduce churn.

PR #59, opened 2016-09-28
Internalize iresponsecachekeyprovider

JunTaoLuo:
cc @BrennanConroy @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/Internal/IResponseCacheKeyProvider.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/Internal/IResponseCacheKeyProvider.cs
@@ -3,9 +3,9 @@
 
 using System.Collections.Generic;
 
-namespace Microsoft.AspNetCore.ResponseCaching
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
 {
-    public interface IResponseCacheKeyProvider
+    internal interface IResponseCacheKeyProvider
     {
         /// <summary>
         /// Create a base key for a response cache entry.
--- a/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheKeyProvider.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheKeyProvider.cs
@@ -10,9 +10,9 @@ using Microsoft.Extensions.ObjectPool;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.Primitives;
 
-namespace Microsoft.AspNetCore.ResponseCaching
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
 {
-    public class ResponseCacheKeyProvider : IResponseCacheKeyProvider
+    internal class ResponseCacheKeyProvider : IResponseCacheKeyProvider
     {
         // Use the record separator for delimiting components of the cache key to avoid possible collisions
         private static readonly char KeyDelimiter = '\x1e';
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheMiddleware.cs
```diff
@@ -32,6 +33,17 @@ public class ResponseCacheMiddleware
             IResponseCacheStore store,
             IOptions<ResponseCacheOptions> options,
             IResponseCachePolicyProvider policyProvider,
+            ObjectPoolProvider poolProvider)
+            :this (next, store, options, policyProvider, new ResponseCacheKeyProvider(poolProvider, options))
+        {
+        }
+
+        // Internal for testing
+        internal ResponseCacheMiddleware(
```
JunTaoLuo: Don't really need this, but kept it to reduce churn in tests.

src/Microsoft.AspNetCore.ResponseCaching/Internal/IResponseCacheKeyProvider.cs
```diff
@@ -3,9 +3,9 @@
 
 using System.Collections.Generic;
 
-namespace Microsoft.AspNetCore.ResponseCaching
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
 {
-    public interface IResponseCacheKeyProvider
+    internal interface IResponseCacheKeyProvider
```
troydai: Should we clear the usage of namespace `*.Internal`? Because I remember not long ago, we decide to use this namespace to hold those types we design to be internal use but still want to offer users the convenience of customization. We left them `public` but keep them in `Internal` namespace so as to indicate that the type may be broken in minor upgrade.

Are we still doing this?

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheKeyProvider.cs
```diff
@@ -10,17 +10,17 @@
...
+    internal class ResponseCacheKeyProvider : IResponseCacheKeyProvider
     {
         // Use the record separator for delimiting components of the cache key to avoid possible collisions
         private static readonly char KeyDelimiter = '\x1e';
 
         private readonly ObjectPool<StringBuilder> _builderPool;
         private readonly ResponseCacheOptions _options;
 
-        public ResponseCacheKeyProvider(ObjectPoolProvider poolProvider, IOptions<ResponseCacheOptions> options)
+        internal ResponseCacheKeyProvider(ObjectPoolProvider poolProvider, IOptions<ResponseCacheOptions> options)
```
troydai: what's the usage of a internal constructor? if the type is already internal what about keep it public?

src/Microsoft.AspNetCore.ResponseCaching/Internal/IResponseCacheKeyProvider.cs
JunTaoLuo: Sort of. I don't think there are a specific set of strict rules for what goes into `*.Internal` namespace. I interpreted it as: these types are internal and users shouldn't need to access them. Some of the types are enforced this way by having the internal accessibility modifier. Other types are still public so that if users really need them, they can use them. Leaving it in the internal namespace adds a big asterisk though, indicating that these types should be used with caution and may change between minor version updates.

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheKeyProvider.cs
JunTaoLuo: It's more of a consistency thing. Although it's not required, retaining a public constructor for an internal type seems inconsistent.

PR #60, opened 2016-09-29
Use HttpMethods

JunTaoLuo:
cc @mikaelm12 @BrennanConroy

### DIFF
--- a/test/Microsoft.AspNetCore.ResponseCaching.Tests/ResponseCachePolicyProviderTests.cs
+++ b/test/Microsoft.AspNetCore.ResponseCaching.Tests/ResponseCachePolicyProviderTests.cs
@@ -11,9 +11,20 @@ namespace Microsoft.AspNetCore.ResponseCaching.Tests
 {
     public class ResponseCachePolicyProviderTests
     {
+        public static TheoryData<string> CacheableMethods
+        {
+            get
+            {
+                return new TheoryData<string>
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.ResponseCaching.Tests/ResponseCachePolicyProviderTests.cs
```diff
@@ -11,26 +11,47 @@ namespace Microsoft.AspNetCore.ResponseCaching.Tests
...
+                    HttpMethods.Get,
+                    HttpMethods.Head
+                };
+            }
+        }
+
         [Theory]
-        [InlineData("GET")]
-        [InlineData("HEAD")]
+        [MemberData(nameof(CacheableMethods))]
```
troydai: can't do inline data?

JunTaoLuo: Nope, the compiler complains that ` error CS0182: An attribute argument must be a constant expression, typeof expression or array creation expression of an attribute parameter type`. The reason being that the definition of these methods are `static readonly` not `const`.

troydai: ok

PR #61, opened 2016-09-29
Make DistributedResponseCacheStore internal

JunTaoLuo:
cc @Tratcher


Tratcher:
I think @DamianEdwards asked to delete this entirely as we already know it won't work for future features.

JunTaoLuo:
So there won't even be a fallback for not supporting more advanced features e.g. tag invalidation?

JunTaoLuo:
Let's consider what other extensibility we want to remove for 1.1. It's possible that CacheEntries and PolicyProvider may also go away?

JunTaoLuo:
:up::date:

PR #63, opened 2016-09-30
StringBuilder extension to uppercase string

BrennanConroy:
Saves `~84` bytes per request
I was getting a range of `~2.6%` to `~8.1%` performance increase. Would like to test it with benchmarks when we add ResponseCaching tests to it.

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/Internal/StringBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/Internal/StringBuilderExtensions.cs
@@ -1,0 +1,22 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
+{
+    internal static class StringBuilderExtensions
+    {
+        internal static StringBuilder AppendUpperInvariant(this StringBuilder builder, string value)
+        {
+            builder.EnsureCapacity(builder.Length + value.Length);
+            for (var i = 0; i < value.Length; i++)
+            {
+                builder.Append(char.ToUpperInvariant(value[i]));
+            }
+
+            return builder;
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheKeyProvider.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCacheKeyProvider.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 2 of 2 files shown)


JunTaoLuo:
What are the perf differences before and after the change? Both memory and processing time.

BrennanConroy:
Before `.387ms - .417ms` per request and `1434` bytes per request
After `.377ms - .386ms` per request and `1350` bytes per request

src/Microsoft.AspNetCore.ResponseCaching/Internal/StringBuilderExtensions.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
+{
+    internal static class StringBuilderExtensions
```
JunTaoLuo: Hmm is this repo a good place to put this? It seems very generic and could be used elsewhere.

src/Microsoft.AspNetCore.ResponseCaching/Internal/StringBuilderExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        internal static StringBuilder AppendUpperInvariant(this StringBuilder builder, string str)
+        {
+            for (var i = 0; i < str.Length; ++i)
+            {
+                builder.Append(char.ToUpperInvariant(str[i]));
```
JunTaoLuo: I thought you had a faster toupper?

JunTaoLuo:
Changes look good, but you'll need to rebase though not sure if the code should live here?

src/Microsoft.AspNetCore.ResponseCaching/Internal/StringBuilderExtensions.cs
BrennanConroy: There was an in-place toupper but that is not safe and only works in very specific scenarios.

BrennanConroy:
Ran with benchmarks and got `1.725%` and `2.778%` increase

src/Microsoft.AspNetCore.ResponseCaching/Internal/StringBuilderExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using System.Text;
+
+namespace Microsoft.AspNetCore.ResponseCaching.Internal
+{
+    internal static class StringBuilderExtensions
+    {
+        internal static StringBuilder AppendUpperInvariant(this StringBuilder builder, string str)
+        {
+            for (var i = 0; i < str.Length; ++i)
```
BrennanConroy: Good idea

pranavkm:
LKGTM

JunTaoLuo:
:shipit:

PR #73, opened 2016-10-13
Updating dependency versions in the sample project

moozzyk:


JunTaoLuo:
Added a minor update for the nuget package verifier. :shipit:

PR #74, opened 2016-10-14
Removing double casts

moozzyk:


.gitignore
```diff
@@ -27,4 +27,5 @@ nuget.exe
 project.lock.json
 /.vs/
 .build/
-.testPublish/
\ No newline at end of file
+.testPublish/
+launchSettings.json
```
Tratcher: please don't ignore these, check them in instead. The settings matter in some of the samples

src/Microsoft.AspNetCore.ResponseCaching/ResponseCacheMiddleware.cs
```diff
@@ -171,24 +177,26 @@ public class ResponseCacheMiddleware
...
                     {
                         return true;
                     }
                 }
             }
-            else if (cacheEntry is CachedResponse && await TryServeCachedResponseAsync(context, (CachedResponse)cacheEntry))
+            else
             {
-                return true;
+                if (await TryServeCachedResponseAsync(context, cacheEntry))
```
JunTaoLuo: ignore this, missed the next little bit during the review

.gitignore
Tratcher: I'd rather we were consistent about checking them in so .gitignore files could be more consistent across repos.

moozzyk: I am not doing this. .gitignores seems to vary by repo. Actually I checked a handful of projects and neither of them had launchSettings.json checked in and some of them already had launchSettings.json in .gitignore. I think .gitignore are typically copied when the repo is created and then they are updated as needed and keeping them consistent is not a goal.

PR #75, opened 2016-10-15
API review

JunTaoLuo:
More API changes:

- ResponseCache* => ResponseCaching* (except store related types)
- Move ResposneCachingOptions to the Microsoft.AspNetCore.ResponseCaching namespace
- Configure options on the AddResponseCaching() services extension
- string[] instead of StringValues for VaryByQueryKeys

cc @Tratcher @glennc


src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs
```diff
@@ -211,7 +210,7 @@ public class ResponseCacheMiddleware
                 // Create the cache entry now
                 var response = context.HttpContext.Response;
                 var varyHeaders = new StringValues(response.Headers.GetCommaSeparatedValues(HeaderNames.Vary));
-                var varyQueryKeys = context.HttpContext.Features.Get<IResponseCacheFeature>()?.VaryByQueryKeys ?? StringValues.Empty;
+                var varyQueryKeys = new StringValues(context.HttpContext.Features.Get<IResponseCachingFeature>()?.VaryByQueryKeys);
```
JunTaoLuo: Converting to `StringValues` here to take advantage of utilities such as `Equals` and `IsNullOrEmpty`. Could consider duplicating these utilities for `string[]` if the extra allocation of a `StringValues` struct is too expensive.

JunTaoLuo:
ping @Tratcher @glennc

src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingFeature.cs
```diff
@@ -2,25 +2,24 @@
...
             {
                 return _varyByQueryKeys;
             }
             set
             {
-                if (value.Count > 1)
+                if (value?.Length > 1)
                 {
-                    for (var i = 0; i < value.Count; i++)
+                    for (var i = 0; i < value?.Length; i++)
```
Tratcher: `?.` is redundant here, you shouldn't reach here without an array.

Issue #81, opened 2016-11-28
Cache-Control: no-cache breaks the middleware


tinchou:
**TL;DR**: this breaks with the `Cache-Control: no-cache` header since the `IResponseCachingFeature` feature is not set up and we hit [this exception](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/ResponseCacheFilter.cs#L132).

-------

Hi,

I'm just starting with your middleware, and the basic functionality is great. I just specify a `[ResponseCache]` and it gets cached automatically, which is awesome :).

**But** I'm getting an error 500 each time I do "Ctrl+F5" or I disable caching in the Chrome dev tools. The error is the one in the title:

```
An unhandled exception occurred while processing the request. InvalidOperationException: 'VaryByQueryKeys' requires the response cache middleware.

Microsoft.AspNetCore.Mvc.Internal.ResponseCacheFilter.OnActionExecuting(ActionExecutingContext context)
```

The error is the same as if I had not called `app.UseResponseCaching();` or called it in the wrong order. The exact line is the following, which shows that it's checking for a `Feature` of type `IResponseCachingFeature`: https://github.com/aspnet/Mvc/blob/d8c6c4ab34e1368c1b071a01fcdcb9e8cc12e110/src/Microsoft.AspNetCore.Mvc.Core/Internal/ResponseCacheFilter.cs#L132.

I investigated the source code and came to the conclusion that the error is produced because:

1. Browsers send `Cache-Control: no-cache` header
1. The request is marked as not cacheable [here](https://github.com/aspnet/ResponseCaching/blob/f8831928de92f7900f26b74f64eb4083227f257a/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingPolicyProvider.cs#L35)
1. Since it's not cacheable we skip the [call to `ShimResponseStream`](https://github.com/aspnet/ResponseCaching/blob/f8831928de92f7900f26b74f64eb4083227f257a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs#L87)
1. And we [don't set up the Feature](https://github.com/aspnet/ResponseCaching/blob/2eb66dc684faeec5085cb957fbe5c4e5a11ec4ec/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs#L340) for the next middlewares.
[...]

PR #82, opened 2016-11-30
Always add IResponseCachingFeature before calling the next middleware

JunTaoLuo:
#81 Will need to see which branch this needs to get into for 1.1.1.

cc @Tratcher


JunTaoLuo:
Changes have been made as part of #88. The fixes will be ported back to 1.1.x in #91.

PR #70, opened 2016-10-07
Header parsing

BrennanConroy:
Still a bit of cleanup needed.

Saved about `791` bytes per request.
`1627.6` originally
`836.6` after

Benchmarks has shown an average of `3.386%` increase in RPS

PR #87, opened 2016-12-12
Optimizations

JunTaoLuo:
While I was looking at the allocation saving on #70 I noticed a few quick optimizations. 

Before: 346747.2 RPS, 1205.3 bytes/req (net451), 848.6 bytes/req (netcore)
After: 434507.1 RPS, 1158.9 bytes/req (net451), 802.8 bytes/req (netcore)

@BrennanConroy @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/Internal/CacheEntry/CachedResponse.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/Internal/CacheEntry/CachedResponse.cs
@@ -13,7 +13,7 @@ namespace Microsoft.AspNetCore.ResponseCaching.Internal
 
         public int StatusCode { get; set; }
 
-        public IHeaderDictionary Headers { get; set; } = new HeaderDictionary();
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.ResponseCaching/Internal/CacheEntry/CachedResponse.cs
```diff
@@ -13,8 +13,10 @@ public class CachedResponse : IResponseCacheEntry
 
         public int StatusCode { get; set; }
 
-        public IHeaderDictionary Headers { get; set; } = new HeaderDictionary();
+        public IHeaderDictionary Headers { get; set; }
 
         public Stream Body { get; set; }
+
+        public string BodyLengthString { get; set; }
```
Tratcher: why not cache this directly in Headers like you do with Date?

JunTaoLuo:
:up::date: @BrennanConroy @Tratcher

Issue #89, opened 2016-12-22
Port fix for Update how certain cache directives are handled to 1.0.x


JunTaoLuo:
The fix for https://github.com/aspnet/ResponseCaching/issues/83 needs to go into LTS 1.0.x branch.

JunTaoLuo:
Need to discuss which parts of the fixes in #88 needs to be back ported to 1.0.x

At the minimum, we'll need the fix for using `If-Modified-Since` instead of the incorrect `If-Unmodified-Since`.

We can potentially also add the handling of proxy-revalidate, and max-stale with no specified limit since those are low risk but they are not critical.

JunTaoLuo:
cc @muratg

muratg:
(Setting milestone to 1.0.1 to reflect the actual package version to be shipped.)

Eilon:
This patch bug is **approved**. Please use the normal code review process w/ a PR and make sure the fix is in the correct branch, then close the bug and mark it as done.

JunTaoLuo:
Patch merged. Will still need to update version numbers.

JunTaoLuo:
Version numbers updated

PR #93, opened 2017-01-21
Use If-Modified-Since instead of the incorrect If-Unmodified-Since header

JunTaoLuo:
Patch PR for #89.

Fixes:
- Using `If-Modified-Since` instead of the incorrect `If-Unmodified-Since` header


BrennanConroy:
:shipit:

PR #115, opened 2017-03-14
Using NullLogger types from Logging.Abstractions

ajaybhargavb:
aspnet/Logging#571
@Eilon


test/Microsoft.AspNetCore.ResponseCaching.Tests/TestUtils.cs
```diff
@@ -15,6 +15,7 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Internal;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
```
Eilon: Is this the only change needed because it was picking up transitive dependencies from somewhere?

ajaybhargavb: Yup. It was already pulling this from somewhere

PR #121, opened 2017-05-04
Migration

JunTaoLuo:


src/Microsoft.AspNetCore.ResponseCaching/Streams/ResponseCachingStream.cs
```diff
@@ -183,19 +183,12 @@ public override void WriteByte(byte value)
             }
         }
 
-#if NETSTANDARD1_3
-        public IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
```
anurse: Is this right? Wouldn't you want this code and not the other one? This branch was the one used by the `netstandard` build (which was actually the only build anyway)

Tratcher: Yes, we need to override Begin/End stream methods. These were not in early standards.

mikeharder:
Merged to dev: https://github.com/aspnet/ResponseCaching/commit/f536c023f08fc42297e703340639b231424023de

mikeharder:
@JunTaoLuo: Please delete branch

PR #123, opened 2017-05-22
Target NETStandard2.0

JunTaoLuo:


samples/ResponseCachingSample/ResponseCachingSample.csproj
```diff
@@ -16,4 +16,7 @@
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' ">
```
pranavkm: `Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework'">`?

JunTaoLuo: We are inconsistent here and from what I see in other repos we use net461 more often.

PR #153, opened 2018-01-17
Add quirk mode to patch fix

JunTaoLuo:
This is for the 1.1 branch

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
@@ -13,14 +13,31 @@ namespace Microsoft.AspNetCore.ResponseCaching.Internal
 {
     public class ResponseCachingKeyProvider : IResponseCachingKeyProvider
     {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
```diff
@@ -13,14 +13,31 @@ namespace Microsoft.AspNetCore.ResponseCaching.Internal
...
+#if NET451
+            var value = System.Configuration.ConfigurationManager.AppSettings.GetValues(UseMethodPathCacheKeySwitch)?.FirstOrDefault();
+            var isEnabled = false;
+            if (bool.TryParse(value, out isEnabled))
+            {
+                UseMethodPathCacheKey = isEnabled;
+            }
+#else
+            var isEnabled = false;
+            AppContext.TryGetSwitch(UseMethodPathCacheKeySwitch, out isEnabled);
```
Tratcher: Can you can use UseMethodPathCacheKey directly?

JunTaoLuo: I tried that and hit `error CS0206: A property or indexer may not be passed as an out or ref parameter`

PR #156, opened 2018-02-22
Delimit QueryString and Header values.

JunTaoLuo:
Addresses https://github.com/aspnet/Home/issues/2810.

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
@@ -15,6 +15,8 @@ namespace Microsoft.AspNetCore.ResponseCaching.Internal
     {
         // Use the record separator for delimiting components of the cache key to avoid possible collisions
         private static readonly char KeyDelimiter = '\x1e';
+        // Use the unit separator for delimiting subcomponents of the cache key to avoid possible collisions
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
```diff
@@ -124,6 +126,10 @@ public string CreateStorageVaryByKey(ResponseCachingContext context)
 
                         for (var j = 0; j < headerValues.Count; j++)
                         {
+                            if (j > 0)
```
JunTaoLuo: Per RFC:

>A recipient MAY combine multiple header fields with the same field
   name into one "field-name: field-value" pair, without changing the
   semantics of the message, by appending each subsequent field value to
   the combined field value in order, separated by a comma.

So the values of the headers will not be delimited.

tangdf:
Query String | Cache Key
--- | ---
?QueryA=ValueA&QueryA=ValueB | Q{KeyDelimiter}QUERYA=ValueA{KeySubDelimiter}ValueB
?QueryA=ValueB&QueryA=ValueA | Q{KeyDelimiter}QUERYA=ValueB{KeySubDelimiter}ValueA

I think they should be the same, both **Q{KeyDelimiter}QUERYA=ValueA{KeySubDelimiter}ValueB**.

JunTaoLuo:
Fair, that seems reasonable. I have opened #157 to sort the header and query values.

PR #157, opened 2018-02-23
Sort header and query values

JunTaoLuo:
Follow up to https://github.com/aspnet/ResponseCaching/pull/156 which addresses https://github.com/aspnet/Home/issues/2810. This might be too much LINQ though. I'll check the performance benchmarks numbers and see if we need to reduce allocations.

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
@@ -82,7 +82,7 @@ namespace Microsoft.AspNetCore.ResponseCaching.Internal
             }
         }
 
-        // BaseKey<delimiter>H<delimiter>HeaderName=HeaderValue<delimiter>Q<delimiter>QueryName=QueryValue
+        // BaseKey<delimiter>H<delimiter>HeaderName=HeaderValue<delimiter>Q<delimiter>QueryName=QueryValue1<subdelimiter>QueryValue2
         public string CreateStorageVaryByKey(ResponseCachingContext context)
         {
             if (context == null)
@@ -124,9 +124,12 @@ namespace Microsoft.AspNetCore.ResponseCaching.Internal
                             .Append(header)
                             .Append("=");
 
-                        for (var j = 0; j < headerValues.Count; j++)
+                        var headerValuesArray = headerValues.ToArray();
+                        Array.Sort(headerValuesArray, StringComparer.Ordinal);
+
+                        for (var j = 0; j < headerValuesArray.Length; j++)
                         {
-                            builder.Append(headerValues[j]);
+                            builder.Append(headerValuesArray[j]);
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


JunTaoLuo:
Before: 458981 RPS After: 270014 RPS, which is about a 40% drop in perf.

JunTaoLuo:
Updated to not use LINQ

Before: 458981 RPS After: 359827 RPS, which is about a 21.6% drop

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
```diff
@@ -141,19 +154,36 @@ public string CreateStorageVaryByKey(ResponseCachingContext context)
...
+                                continue;
+                            }
+                            if (queryArray[i].Value.Count == 1)
+                            {
+                                builder.Append(queryArray[i].Value[0]);
+                                continue;
+                            }
+
+                            var queryValueArray = queryArray[i].Value.ToArray();
+                            Array.Sort(queryValueArray, StringComparer.Ordinal);
```
JunTaoLuo: Note to self: check if Ordinal comparison is appropriate.

JunTaoLuo:
Ah, I've been running the benchmarks in Debug configuration. Using the Release configuration I'm getting 479127 RPS which is higher (?) than before the change, likely due to noise. Regardless, this suggests that the change has minimal impact on perf.

sebastienros:
@davidfowl you can even benchmark debug assemblies with this technique, how convenient ;)
Maybe a warning would not hurt if there is a way to detect it, and an assembly ...

@JunTaoLuo to make noise smaller you can run a longer benchmark, with `--duration 60` for instance. You can also repeat it and get an average result like `-i 3` to run the same task 3 times. But my preference goes for duration.

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
```diff
@@ -141,19 +144,26 @@ public string CreateStorageVaryByKey(ResponseCachingContext context)
                     if (varyByRules.QueryKeys.Count == 1 && string.Equals(varyByRules.QueryKeys[0], "*", StringComparison.Ordinal))
                     {
                         // Vary by all available query keys
-                        foreach (var query in context.HttpContext.Request.Query.OrderBy(q => q.Key, StringComparer.OrdinalIgnoreCase))
+                        var queryArray = context.HttpContext.Request.Query.ToArray();
+                        Array.Sort(queryArray, QueryKeyComparer.Ordinal);
```
JunTaoLuo: Might not need to do this

Tratcher: IgnoreCase? Check how query aggregates

JunTaoLuo:
:up::date:

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
```diff
@@ -141,19 +144,26 @@ public string CreateStorageVaryByKey(ResponseCachingContext context)
                     if (varyByRules.QueryKeys.Count == 1 && string.Equals(varyByRules.QueryKeys[0], "*", StringComparison.Ordinal))
                     {
                         // Vary by all available query keys
-                        foreach (var query in context.HttpContext.Request.Query.OrderBy(q => q.Key, StringComparer.OrdinalIgnoreCase))
+                        var queryArray = context.HttpContext.Request.Query.ToArray();
+                        Array.Sort(queryArray, QueryKeyComparer.OrdinalIgnoreCase);
```
Tratcher: Comment this one is ignore case because Request.Query aggregates keys using ignore case wherease the values are compared ordinally.

src/Microsoft.AspNetCore.ResponseCaching/Internal/ResponseCachingKeyProvider.cs
```diff
@@ -124,9 +124,12 @@ public string CreateStorageVaryByKey(ResponseCachingContext context)
                             .Append(header)
                             .Append("=");
 
-                        for (var j = 0; j < headerValues.Count; j++)
+                        var headerValuesArray = headerValues.ToArray();
+                        Array.Sort(headerValuesArray, StringComparer.Ordinal);
```
jkotalik: Anyway we can save the sorted headerValues array if they haven't been modified? Though don't think it will matter much because CreateStorageVaryByKey is called usually once per request.

jkotalik: Also can we check if the size is 1 before we sort? Idk if it will have meaningful perf impact.

JunTaoLuo: Right it's called per request and per cached response so saving it won't be useful. I have tried checking the size and special casing 0 or 1 element. I didn't notice any performance difference and reverted the change to keep things clean. Seems like ToArray and Array.Sort handle these cases pretty well.

PR #172, opened 2018-07-26
We don't use the IMemoryCache from DI

JunTaoLuo:
Instead we create a private instance: https://github.com/aspnet/ResponseCaching/blob/master/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingMiddleware.cs#L40-L43

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingServicesExtensions.cs
+++ b/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingServicesExtensions.cs
@@ -25,7 +25,6 @@ namespace Microsoft.Extensions.DependencyInjection
                 throw new ArgumentNullException(nameof(services));
             }
 
-            services.AddMemoryCache();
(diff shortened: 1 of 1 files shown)


Tratcher:
Careful. Anybody that was resolving that memory cache can break if you remove it. 

Is that why you're doing it only in Master?

Eilon:
Yeah this is a breaking change but I think a good one. I'd say it's fine for 3.0.

JunTaoLuo:
Right, it's a breaking change but we shouldn't be adding services we don't use.

Eilon:
I'm fine leaving 2.2 as-is, and fix only in 3.0. Mark the bug as breaking-change. (No need for an announcement unless someone complains.)

Eilon:
Actually, *is* there an issue for this? If not, please log one.

JunTaoLuo:
No there wasn't. Logged at: https://github.com/aspnet/Home/issues/3360.

